PR #2, opened 2016-02-18
Fix several memory leaks and reference counting snafus in the C++ code.

jamesjer:
I maintain the Fedora gnofract4d package.  We have had a number of crashes reported to us, all apparently due to bugs in the C++ code.  This patch fixes all of the issues I was able to find.  I believe that the reasons for the changes are fairly obvious, but let me know if something in here is mysterious.

### DIFF
--- a/fract4d/c/fract4dmodule.cpp
+++ b/fract4d/c/fract4dmodule.cpp
@@ -226,12 +226,14 @@ get_double_array(PyObject *pyitem, const char *name, double *pVal, int n)
     if(!PySequence_Check(pyfield))
     {
 	PyErr_SetString(PyExc_ValueError, "Bad segment object");
+	Py_DECREF(pyfield);
 	return NULL;
     }
 
     if(!(PySequence_Size(pyfield) == n))
     {
 	PyErr_SetString(PyExc_ValueError, "Bad segment object");
+	Py_DECREF(pyfield);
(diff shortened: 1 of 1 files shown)


edyoung:
Thanks Jerry. Cool, I didn't actually realize there was an active Fedora package. I'll take a look this weekend.

edyoung:
There was a bug in your change (you had != instead of == at one point) which meant that the program quit on startup. With that fixed, have merged the rest. 

Not sure these can account for reports from end users of crashes, most are on error paths I'd not expect to bit hit under normal circumstances - so there may still be issues out there. 

Thanks for the change.

jamesjer:
Maybe so.  We'll see if the crash reports tail off or not. :-)  Thanks for looking at it.

edyoung:
Are these automated crash reports or manual? Where would I see them?

jamesjer:
They are automated.  You can find them on https://retrace.fedoraproject.org/faf/.  Type "gnofract4d" into the "Components" box.  The main page, "Summary", shows bug counts by product (Fedora 22, Fedora 23, EPEL 7, etc.).  The "Reports" tab will show you the actual crash reports.  You can also view information about the Fedora gnofract4d package by visiting https://bugz.fedoraproject.org/gnofract4d.

PR #10, opened 2018-01-23
[WIP] More Python 3 / GTK+ 3 Work

cjmayo:
I've had a go at progressing the Python 3 branch.

My immediate goal was to get the gnofract4d executable to start - I've got that far. I could do more GTK+ conversion but the big blocker is getting the fractal to render. I don't really understand how the existing code works so can't make more progress.

A couple of things I have just disabled for now: stock buttons and Flickr support (see commit messages).
And the last commit is definitely unfinished and not ready.
Anyway all is worth reviewing before committing.

N.B. I've always got:
 ```
./setup.py build
./gnofract4d

Can't import a required module.
If you haven't set up Gnofract 4D yet, run './setup.py build'
Error was: 'cannot import name 'fract4dc''
```
so have never managed to run the tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ doc/gnofract4d-manual/C/gnofract4d-manual.pdf
 doc/gnofract4d-manual/C/stdlib.xml
 elephant-valley/*
 experiments/*
+fract4d/c/cmap_name.h
 images/*
 other_maps/*
 params/*
--- a/benchfct.py
+++ b/benchfct.py
@@ -13,7 +13,7 @@ class Benchmark:
 
     def build(self):
         cmd = self.command + " --buildonly foo.so " + self.args
-        print cmd
+        print(cmd)
         result = os.system(cmd)
         assert result == 0
 
@@ -36,13 +36,13 @@ for arg in sys.argv[1:]:
 bench = Benchmark(command, args)
 bench.build()
 times = []
-for i in xrange(repeats):
+for i in range(repeats):
     t = bench.run()
-    print t
+    print(t)
     times.append(t)
 
-print "stats"
-print "min:(%.4f), max(%.4f), average(%.4f)" % (min(times), max(times), sum(times)/len(times))
+print("stats")
+print("min:(%.4f), max(%.4f), average(%.4f)" % (min(times), max(times), sum(times)/len(times)))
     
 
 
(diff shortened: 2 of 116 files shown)


cjmayo:
OK. Now with pictures. The main trick is not to call `redraw_rect` directly but request it with `widget.queue_draw_area`, a Cairo context gets sent automatically as part of the draw event and `Gdk.cairo_set_source_pixbuf` can write a GdkPixbuf to that Cairo context. Looks like the images are straight RGB with no alpha.

Still more to do, notably dialogs and stock icons but looks possible now.

cjmayo:
Memory problems worked around by passing `pixbuf.copy()` to Gdk.cairo_set_source_pixbuf. Idea from:
https://stackoverflow.com/a/32586739
I tried `del pixbuf; gc.collect()` (which I have used in another program) but it didn't seem to work.

Also added the selection rectangle by storing coordinates in a variable and then drawing in redraw_rect.

cjmayo:
Fixed document generation with Python 3.

Some fixes for things already in my branch. Can squash them in later.

cjmayo:
Added to and rebased. I've squashed quite a few together but probably could do more (and even more with the original python3 branch).

Dialogs and stock icons are working now, as is running gnofract4d in the source directory after`./setupy build`.

`test.py`'s and `test_*`'s work from within their own directories. Lots of deprecation warnings still and a few test failures in fract4dgui.

Three commits disabling things:
disable fractal settings colors page - drawing needs to be converted to Cairo and also colormap-handling functions of GtkWidget have been removed. Maybe create something like gtkfractal.count_colors() instead?
disable failing testColossalImage
disable Flickr features in main_window

Don't understand why this one is necessary or if it is a problem:
[?] fix pan fourway not initialised properly

edyoung:
Although I haven't done a ton of testing, this is clearly way further ahead than what's in python3, so I've gone ahead and merged it in. Thanks again!

edyoung:
fyi i fixed a bunch of deprecation warnings and deleted flickr support in the python3 branch, and added a few issues to the issue tracker to cover things which I think need fixing before merging python3 branch to master. Nearly there, though...

cjmayo:
Thanks for the heads-up.

I'll monitor the Atom feed of commits!

PR #23, opened 2018-02-12
Alert Message Dialog

cjmayo:
Excess space above message and below buttons. Resolved by simplifying the code using Gtk.MessageDialog built-in features.

Also ignore_info no longer used after Flickr code removed, suggest deleting.


edyoung:
Thanks a lot for all the changes Chris - really nice to be making some progress!

PR #24, opened 2018-02-15
Dialog Objects

cjmayo:
I'd seen various bits of odd behaviour, first with the fourway's leading to a89e39cb8a8990e12ef674f6ed3d24363ece0ef0.
Also GTK+ Critical errors on the command line and Python exceptions. 

And something strange in the code:
https://github.com/edyoung/gnofract4d/blob/572dbeecde9cef266167a479368675c710e77926/fract4dgui/browser.py#L162-L163

I can't find any other examples of testing the existence of GtkTreeSelection, indeed the GTK+ docs say:
"The GtkTreeSelection object is automatically created when a new GtkTreeView widget is created, and cannot exist independently of this widget."

I think this is all the result of a confusion of classes, objects, static methods and scope. This set attempts to fix that.

There are a couple of improvements too:
- The formula browser now opens with the current file and formula selected
- The fractal settings can now be dragged wider by the user (but not shrunk).

Hope that makes sense. There a few comments in the commit messages.

### DIFF
--- a/fract4d/browser_model.py
+++ b/fract4d/browser_model.py
@@ -99,12 +99,12 @@ class T:
             TypeInfo(self, compiler, fc.FormulaTypes.TRANSFORM),
             TypeInfo(self, compiler, fc.FormulaTypes.GRADIENT)
             ]
+        self.current = None
         self.current_type = -1
         self.type_changed = event.T()
         self.file_changed = event.T()
         self.formula_changed = event.T()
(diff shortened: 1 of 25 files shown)


cjmayo:
Rebased to remove this workaround:
https://github.com/edyoung/gnofract4d/blob/88617955e70793e9ffe496553eeecf4ce5d1bec5/fract4d/cache.py#L36
that is fixed here.

PR #25, opened 2018-02-19
Fix Video Creation

### LINKED ISSUE #7: Director mode not working.
First of all, great software!
The issue I'm reporting is related to the director mode:
I've imagemagick and transcode installed and compiled, but every time i try to render a video it ends up with nothing:
I can see the render screen working, but when it reaches the end nothing happend.
The thing is that i've tryied to modify the directory for tmp images and i can see that pictures are rendered but when all are done the program doesn't produce any video.
Also i cannot see any output file called like "render.avi" but instead only an "animation.fcta".
Plus i cannot write inside the space to select a name to my new animation, i can just select a file from "browse", and even creating an empty file called "animation.avi" and selecting it i end up with nothing but a bunch of pictures.

cjmayo:
This is on top of #24.

Hopefully fixes #7, transcode option `--use_rgb` is not supported in 1.1.7 and problems with the image list too. As latest transcode is I believe 7 years old and looks like it won't work with ImageMagick 7 I've changed the code to use ffmpeg direct. The video is set to webm format.

I removed the threading and used GLib.spawn_async instead. As well as simplifying the code a plus is that it can now detect success/failure and the user can cancel generation. I haven't implemented progress reporting though.

Added the ability to type the video filename straight into the Directory dialog.

Fixed a bug in that all the settings in the Director dialog were set to 0 on opening.

Usual tidy up and fix as many deprecation warnings as possible.

This does generate video but worth checking it is as expected.

The ffmpeg file list format can support durations for each listed frame so some savings may be possible by using that.

### DIFF
--- a/doc/gnofract4d-manual/C/gnofract4d-manual.xml
+++ b/doc/gnofract4d-manual/C/gnofract4d-manual.xml
@@ -354,7 +354,7 @@ Tips:
 
 <listitem><para>
 In order to end up with a video file, not just a bunch of images, you need to have
-<emphasis>ffmpeg</emphasis> compiled with support for zlib.
+<emphasis>ffmpeg</emphasis> compiled with support for zlib and libvpx.
 </para></listitem>
 
 <listitem><para>
--- a/doc/gnofract4d-manual/C/gnofract4d-manual.xml
+++ b/doc/gnofract4d-manual/C/gnofract4d-manual.xml
@@ -345,17 +345,16 @@ keyframe will stay in video (<guilabel>stopped for</guilabel>) and
 interpolation type between several possibilities. When you hit
 <guibutton>Render</guibutton> button, Director will render all frames
 and put them in directory you selected and then it will try to create
-video using <ulink
-url="http://www.transcoding.org/"><emphasis>transcode</emphasis></ulink>
-tool.  </para>
+a video using <ulink url="https://www.ffmpeg.org/">
+<emphasis>FFmpeg</emphasis></ulink>.</para>
 
 <para>
 Tips:
 <itemizedlist>
 
 <listitem><para>
-In order to end up with video file, not just a bunch of images, you need to have
-<emphasis>transcode</emphasis> tool compiled with support for ImageMagick.
+In order to end up with a video file, not just a bunch of images, you need to have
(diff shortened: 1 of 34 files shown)


cjmayo:
Video uses VP9, mention that in docs and commit message.

edyoung:
Sorry for delay, I have been out of town. I have added you to the collaborators list, please feel free to make any changes to python3 branch. Hopefully we can merge that back to master fairly soon...

cjmayo:
No hurry. I've got another batch on the go - I think that will then be it for the minimum needed to switch over to Python 3 - assuming the selection box performance is acceptable now. Even a few minor improvements included!. Just need a bit more time to tidy it up.

Thanks for the invite.

cjmayo:
I've pushed to the python3 branch. That is all I have (except something on the icons which I will post on that issue) - which is just as well because it is quite a lot. N.B. to review especially: I have removed unused code from both fract4d and frac4dgui, saving what looked like it was stand-alone to fractutils. Also removed fractutils from installation in setup.py, some of it still needs conversion and I think at least the executables need to go somewhere other than site-packages. I'll summarise some of the more interesting items to try and make it a bit easier to digest:

User visible

Additions:

Main window size is saved on exit and restored on startup
Explorer menu and toolbar items work together
Option to set the cache directory
Add filter for video file
Queue dialog automatically hides when render complete

Changes:

Don't install fractutils. Conversion to Python 3 is not complete and installing executables in site-packages is not very convenient.
Updated README for requirements

Fixes:

Save as hi-res was overwriting Save as current dialog
Allow director dialog to be hidden and then be reshown with current settings (needed to add multiple keyframes from current fractal)
Restore display size on return from full screen mode


Developer

Changes:

Rewrite comand-line options with argparse
Write all test output to temporary directory
async parameter renamed to asynchronous because async will be a reserved word in Python 3.7
Unused Python scripts removed from fract4d and fract4dgui, if executable moved to fractutils.
gnofract4d now creates a fractconfig.T object and passes it to MainWindow or fractmain.T  instead of it being a side effect of importing fractconfig.
GUI user preferences are now an object belonging to MainWindow instead of a side effect of importing preferences.

Fixes:
Get profiling working with Python 3

Issue #34, opened 2018-11-20
Unable to setup: NO PNG HEADERS FOUND, you need to install libpng-dev


canpoyrazoglu:
I'm on a fresh Ubuntu 18.10 installation (just installed on a fresh VM).

When I try to build using "./setup.py build" I get the following error:

```
can@ubuntu:~/Desktop/gnofract4d$ ./setup.py build
Can't find 'libpng'
Some functionality will be disabled
Traceback (most recent call last):
  File "./setup.py", line 53, in <module>
    raise Exception("NO PNG HEADERS FOUND, you need to install libpng-dev")
Exception: NO PNG HEADERS FOUND, you need to install libpng-dev
```

I've installed `libpng`: 

```
can@ubuntu:~/Desktop/gnofract4d$ sudo apt-get install libpng-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following additional packages will be installed:
  libpng-tools
The following NEW packages will be installed:
  libpng-dev libpng-tools
0 upgraded, 2 newly installed, 0 to remove and 21 not upgraded.
Need to get 202 kB of archives.
After this operation, 703 kB of additional disk space will be used.
Do you want to continue? [Y/n] 
Get:1 http://us.archive.ubuntu.com/ubuntu cosmic/main amd64 libpng-dev amd64 1.6.34-2 [177 kB]
Get:2 http://us.archive.ubuntu.com/ubuntu cosmic/main amd64 libpng-tools amd64 1.6.34-2 [25.6 kB]
Fetched 202 kB in 1s (169 kB/s)         
Selecting previously unselected package libpng-dev:amd64.
(Reading database ... 171360 files and directories currently installed.)
Preparing to unpack .../libpng-dev_1.6.34-2_amd64.deb ...
Unpacking libpng-dev:amd64 (1.6.34-2) ...
Selecting previously unselected package libpng-tools.
Preparing to unpack .../libpng-tools_1.6.34-2_amd64.deb ...
Unpacking libpng-tools (1.6.34-2) ...
Setting up libpng-tools (1.6.34-2) ...
Processing triggers for man-db (2.8.4-2) ...
Setting up libpng-dev:amd64 (1.6.34-2) ...


```

However, I'm still getting the exact same error about not finding libpng. What am I doing wrong? Why can't it find `libpng` after installing and why can't it build properly as readme clearly states that libpng is optional anyway?

edyoung:
Please see if this PR addresses the issue

[...]

PR #36, opened 2018-11-21
Improved checks for Python and pkg-config

cjmayo:
Updates to setup.py inspired by issues reported in #34 and some tidy-ups.

Some notes in commit messages.

Tested on Gentoo with Python 3.6. I believe it will still run on Python 3.5.


edyoung:
Sorry, forgot to look at this. Thanks!

PR #45, opened 2019-05-21
improve video generation

cjmayo:
Speed up by removing Python threading and solely using the signals emitted by gtkfractal.

Clarify that a video is made up of keyframes and transitions consisting of a number of frames defined by the duration parameter. By default the keyframe is shown for one frame but this is configurable using the stopped parameter.

---

Only the last commit is the new video work, all the others could go straight in. (it is based on the branch in #43).

One last issue, present in 4.0.1 too - does the blend function work? The sequence of transition frames seems to stop well short of the next keyframe?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,6 @@ env: PIP_NO_BUILD_ISOLATION=false
 install:
   - python setup.py build
   - pip install tox-travis
-script: xvfb-run tox
+script: tox
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
   - "3.7"
 before_install:
   sudo apt-get install -y libgirepository1.0-dev gir1.2-gtk-3.0
+env: PIP_NO_BUILD_ISOLATION=false
 install:
   - python setup.py build
   - pip install tox-travis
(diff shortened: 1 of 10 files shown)


edyoung:
Would if be best to rebase ?

cjmayo:
Rebased on master. Travis isn't working for me with `xvfb-run` locally either...

cjmayo:
Still failing even with XAUTHORITY. I can't see the Travis logs so I'll stop guessing.

cjmayo:
I had some of the Travis JavaScript blocked which was why I couldn't see the logs. I've put the necessary fixes in #47.
Then split the video fixes into #48.
This one is now based on top of both of those and just has the video changes.

Interestingly we are occasionally hitting a failure not of these changes but fract4d/test_fract4d.py

```py
self = <fract4d.test_fract4d.Test testMethod=testFDSite>
...
                nb = 2*4
                bytes = os.read(rfd,nb)
                if len(bytes) < nb:
>                   self.fail("bad message with length %s, value %s" % (len(bytes), bytes))
E                   AssertionError: bad message with length 4, value b'\x03\x00\x00\x00'
fract4d/test_fract4d.py:573: AssertionError

PR #52, opened 2020-01-30
Fix fract4d/test_image.py

josebailo:
The test provokes an error and checks the message.
If you don't have the system in english the test fails.
Now the test only checks part of the message text so the
language shouldn't be a problem.


cjmayo:
Really minor point but would `startswith` and `endswith` work instead of `find`?
(If not then perhaps `in`)

PR #53, opened 2020-01-30
Fix macos compatibility

mindhells:
There were some errors on fractconfig implemenation preventing the compiler to get the right flags for MacOS system.
Also some tests wasn't checking the environment to create the userConfig object.
Finally there was a problem with codegen (a module used for the compiler) test: it was looking for "-nan" values but, depending on the system compiler, those values could be "nan" or "-nan".

### DIFF
--- a/fract4d/test_codegen.py
+++ b/fract4d/test_codegen.py
@@ -232,7 +232,11 @@ int main()
         cFile.flush()
 
         oFileName = str(Path(cFile.name).with_suffix(".so"))
-        cmd = "gcc -Wall -fPIC -DPIC -shared %s -o %s -lm" % (cFile.name, oFileName)
+        import sys
+        if sys.platform[:6] == "darwin":
+            cmd = "gcc -Wall -fPIC -DPIC -shared %s -o %s -lm -flat_namespace -undefined suppress" % (cFile.name, oFileName)
+        else:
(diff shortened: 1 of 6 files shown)


fract4d/test_codegen.py
```diff
@@ -1768,7 +1772,7 @@ def mycotanh(z):
         #print tests
         
         # CONSIDER: Python 2.7 thinks this cotanh(0) is -nan,-nan, older versions think (nan,nan)
-        tests[0][2] = "(-nan,-nan)"
```
edyoung: Since we know that some environments produce -nan and some produce nan, is it possible to adjust the test logic so that we would accept either? 
 think in general the distinction between -nan and nan is pretty academic, so if we could make it so either was accepted throughout, that would be ideal.

edyoung:
Thanks! Other changes look fine. LMK what you think on the test comment

fract4d/test_codegen.py
mindhells: That nan/-nan value comes from C++, and that's why I changed the C++ code produced by  `inspect_complex` function to make sure we always get "nan".
This functions `cotanhtests` and `cotantests` are called by the test `test_stdlib` which at the end relies on the mentioned ìnspect_complex` to generate the C++ that is compiled and executed to check the return values.

PR #59, opened 2020-03-07
Run tests on Python 3.8

cjmayo:
Working OK for me. Let's see how Travis does...


cjmayo:
Needs fixing. Looks like pycairo needs to be installed before it tries to build pygobject.

cjmayo:
Working now, although I did have to update the image from Xenial to Bionic.

PR #60, opened 2020-03-07
Don't install tests

cjmayo:
Not needed by ordinary users, plus test parameter files in testdata are
not installed.

---

Something I had wanted to do for a long time, now straightforward thanks to the work of @guanchor.

Plus fix SyntaxWarnings visible with Python 3.8.


edyoung:
thanks!

PR #62, opened 2020-03-16
Refactor fract4dc

mindhells:
This is a first step towards refactoring the fract4dc python C++ extension sources.
It's very hard to maintain or add features to this part of the codebase because of the naming, formatting and organization of the code.

`fract4dmodule.cpp` file is way too large for it's purpose which should be only to expose the extension interface to Python. For that reason, only the static functions and the python module definition itself had been preserved, all the implementations have been organized into subjects and moved to a `fract4dc` subfolder. In that `fract4dc` subfolder, .h and .cpp file pairs have been created for every subject (sites, images, calcs...). Each of those pairs define a namespace with the same name and contain the implementation that was in `fract4dmodule.cpp`.

Additionally, some dead code has been removed:
- the creation of fract4d_stdlib extension in the setup.py is not needed therefore has been removed 
- all references and files regarding windows support had been deleted as well
- some unused and unmaintained cpp files had been removed


mindhells:
need to fix test 1st

PR #65, opened 2020-03-18
Refactor fract4dc

mindhells:
This is a first step towards refactoring the fract4dc python C++ extension sources.
It's very hard to maintain or add features to this part of the codebase because of the naming, formatting and organization of the code.

`fract4dmodule.cpp` file is way too large for it's purpose which should be only to expose the extension interface to Python. For that reason, only the static functions and the python module definition itself had been preserved, all the implementations have been organized into subjects and moved to a `fract4dc` subfolder. In that `fract4dc` subfolder, .h and .cpp file pairs have been created for every subject (sites, images, calcs...). Each of those pairs define a namespace with the same name and contain the implementation that was in `fract4dmodule.cpp`.

Additionally, some dead code has been removed:
- all references and files regarding windows support had been deleted as well
- some unused and unmaintained cpp files had been removed

My initial idea was to remove the fract4d_stdlib extension from setup and fract4dc module but I found that in some environments (like travis runner), and because of how the dynamic linking works on them, this libs needs to be loaded as Global `dlopen(fract4d_stdlib_path, RTLD_GLOBAL | RTLD_NOW)`. 
There are some workarounds to solve that but I'm not sure which environments are affected. I can tell that you can just remove the extension on MacOS.
The reason for removing this fract4d_stdlib extension is its sources are already included in the fract4dc module.

To figure out the problem with the stdlib I had to reproduce locally the travis environment. For that I created a docker image. I've included an utility script using docker-compose to load that image into a container and pass the tests. It's under `docker/bin/run_tests.sh`.


edyoung:
I remember initially adding stdlib.so because of the reason you mention, but I have no idea which platforms forced this.

Thanks for tidying up!

edyoung:
The 'arena' functions were a first step toward implementing array support in the fractal compiler (ultrafractal has this). In general deleting unused code makes sense but those particular parts are fairly close to working if I recall correctly so if there's any interest in expanding the language in that direction may be worth keeping them

mindhells:
You're welcome.
There's more tidying coming for the next week! (If you are courious you can check [the progress](https://github.com/HyveInnovate/gnofract4d/tree/refactor-fract4dc-phase2))

Btw I got an email with a comment from you regarding `arenas`. I think you removed it because you 1st though that I removed that part but I didn't. You mentioned that was intended to implement `array support`, a feature included in Ultrafractal. Can you please explain what's that about?

edyoung:
The formula language in Gnofract 4D is a superset of the language
originally introduced in Fractint, but a subset of the language in
UltraFractal. UF allows you to declare array variables and use them in your
fractals. I was planning to add support for that. See
https://www.ultrafractal.com/help/index.html?/help/writing/language/arrays.html

mindhells:
Hi @edyoung,
is there something you need to continue with merging this PR?
I have more incoming based in this status. Please let me know what are your plans.

edyoung:
I approved it previously to let you merge when you felt ready. Will do it now

edyoung:
Feel free to delete it

guanchor:
I have some missing permissions to achieve it :-)

PR #66, opened 2020-03-31
Fix several errors in Python

guanchor:
Since we have pylint integrated in travis building phase for PEP8 rules validation in the Python code, i have activated also the errors flag for pylint.

All the issues were fixed but false positive about external modules as **gi** and **cairo**.


edyoung:
Seems like Travis tests for Python 3.8 are crashing. Presumably these changes are unrelated but would be great it we can fix that to get a reliable CI build...

edyoung:
merged https://github.com/edyoung/gnofract4d/pull/69 which had the same changes, after resolving merge conflict. Thanks!

edyoung:
@guanchor  A couple notes on this. When I run bin/pylint.sh on my system (ubuntu 18.04) it tries to run pylint for python2 which complains about python3 syntax. If I install pylint3 and change the script to run that, it works better. But I still get a number of errors reported:

```
+-----------------------+------------+
|message id             |occurrences |
+=======================+============+
|no-member              |316         |
+-----------------------+------------+
|used-before-assignment |5           |
+-----------------------+------------+
|import-error           |1           |
+-----------------------+------------+
|catching-non-exception |1           |
+-----------------------+------------+

```

Not sure why, any ideas?

guanchor:
@edyoung I couldn't fix the conflicts past week, but I'm glad you did it without any complications.

> When I run bin/pylint.sh on my system (ubuntu 18.04) it tries to run pylint for python2 which complains about python3 syntax. If I install pylint3 and change the script to run that, it works better.

Looks like your python alias points to python2 instead of python3. You can swap this or we can change the line in the pylint.sh.

In my opinion is better to use the docker that we have now to try to execute locally the same pipeline actions with the same configuration than travis.

> But I still get a number of errors reported

- Can you check you are really executing pylint with the pylintrc configuration?
- You are using python3, but what about pylint? I am using pylint2.4.4. There are a lot of false positive that are being fixed in the pylint project, so it could be nice to know your version so we can update the dependecies.

Thanks!!

guanchor:
@edyoung you should remove your current pylint and re-install pylint throw pip3 (not pip) to ensure that pylint is running for Python3

PR #67, opened 2020-04-03
refactor fract4dc phase 2

mindhells:
In this second phase the python extension specific code (for dealing with inputs and outputs with the python program) and the core business logic of the extension has been separated. This could make sense in the future to create a new project only for a C++ library, for now it makes easy maintenance and improvement.

All the modules supporting the python interface are located now under `fract4dc` directory. The rest of the code has been moved to `model` directory, with some exceptions:
- `pf.h`: this file remains untouched because it's copied into the formula sources by the compiler
- `threadpool.h` and `vectors.h`: those two are modified versions of what should be external dependencies.
- `fract_stdlib.(cpp|h)` and `colorutils.(cpp|h)`: those are part of the second extension being compiled in the setup.py. 

After all this reorganization, [cppclean tool](https://github.com/myint/cppclean) has been used to revise and clean all the includes. Now all dependencies between modules can be easily identified.

`Benchmark.py` script has been fixed to compare results before and after the refactor: no noticeable differences has been observed.

There's some pending points for a new release:
- fract_stdlib extension: while it cannot be removed it would be nice at least remove the need to include `model/image.cpp` among the sources needed to compile it.
- inside `model`: there's still 2 entities that are meant to be used only by the python interface `PySite` and `calcargs`. Maybe it makes sense just to move them inside `fract4dc` directory.
- move all sources outside {projectDir}/fract4d


fract4d/c/model/fractfunc.h
```diff
@@ -1,118 +1,123 @@
...
+    dvec4 eye_point;              // where user's eye is (for 3d mode)
 
- private:
+private:
     // MEMBER VARS
-
     bool ok; // did this instance get constructed ok?
-    // (* this should really be done with exns but they are unreliable 
+    // (* this should really be done with exns but they are unreliable
     //  * in the presence of pthreads - grrr *)
```
edyoung, line 102: btw this comment was true in ~2003 but hopefully isn't any more

edyoung:
Applied these changes via #71 . Thanks a lot!

PR #85, opened 2020-04-21
New fract4dc interface

mindhells:
A new interface for the fract4dc has been created by introducing a new [python extension type](https://docs.python.org/3/extending/newtypes_tutorial.html) called `controller`.
The goal of this new interface is to avoid objects swinging in and out from fract4dc to python and the other way around all the time (which forces the need of having ugly interfaces and complicates the state handling).
No python scripts have been modified yet but only a new test has been created to demonstrate how to use it. Also the methods that would be deprecated by moving to this new interface have been marked in `fract4d_module.cpp`
In addition, a new folder [examples](examples) has been created containing:
- python examples: demonstrating the minimum steps necessary to generate a mandelbrot image (using the new interface)
- C++ standalone examples: to explain how the engine (the pure C++ part of the fract4dc extension) works to do the same as the example above

The idea behind all this is to transfer all the knowledge gotten from analyzing and reorganizing the code to future developers.


fract4d/c/fract4dmodule.cpp
```diff
@@ -413,7 +422,145 @@ pyarray_set(PyObject *self, PyObject *args)
...
 
 
+/*
+* Controller extension type
+*/
+
+static PyObject *
+Controller_set_message_handler(FractalController *self, PyObject *args)
+{
+    PyObject *message_hanlder;
```
edyoung, line 432: typo in name, handler

examples/python/simple_mandelbrot.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python3
```
edyoung, line 1: and a gui-free python version too!

examples/cpp/multithread_mandelbrot.cpp
```diff
@@ -0,0 +1,184 @@
+#include <stdlib.h>
```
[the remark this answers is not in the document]
mindhells: that's true. In this example though there's a function reading that output, just to illustrate you can be posted about the progress through a pipe using FDSite.
In gnofract4d you can see this process of watching and read the pipe in `gtkfractal.py`

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
+#include <stdlib.h>
```
DX-MON, line 1: Prefer `#include <cstdlib>` - being able to use the .h form from C is incidental behaviour that may be patched out the standard or warned upon in the future. Tools such as `clang-tidy` already issue a warning about this.

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <dlfcn.h>
+#include <stdio.h>
```
DX-MON, line 3: Same comment as stdlib.h

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include "model/imagewriter.h"
+
+#include "model/fractfunc.h"
+#include "model/vectors.h"
+
+
+int main() {
+    // initial setup: load fract4_stdlib globally so the loaded formula has access to it
+    void *fract_stdlib_handle = dlopen("./fract_stdlib.so", RTLD_GLOBAL | RTLD_NOW);
+    if (NULL == fract_stdlib_handle) {
```
DX-MON, line 22: prefer `nullptr`. NULL is a C macro, nullptr is a considerably more robust way of expressing this.

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
...
+        0.0, 0.0, 0.0, 0.0, // X Y Z W
+        4.0, // Size or zoom
+        0.0, 0.0, 0.0, 0.0, 0.0, 0.0 // XY XZ XW YZ YW ZW planes (4D stuff)
+    };
+
+    // initialize the point function with the params
+    pf_handle->vtbl->init(pf_handle, pos_params, params, param_len);
+
+    // create the colormap with 3 colors
+    ListColorMap *cmap = new (std::nothrow) ListColorMap();
```
DX-MON, line 72: Use std::unique_ptr<> - This could be rewritten to `std::unique_ptr<ListColorMap> cmap{new (std::nothrow) ListColorMap{}};` without impacting intent, while then gaining the benefits of automatic memory management

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
...
+    params[3].t = FLOAT;
+    params[3].doubleval = 1.0;
+    params[4].t = FLOAT;
+    params[4].doubleval = 4.0;
+    params[5].t = FLOAT;
+    params[5].doubleval = 0.0;
+    params[6].t = FLOAT;
+    params[6].doubleval = 1.0;
+    // position params
+    double pos_params[N_PARAMS] {
```
DX-MON, line 62: This could be constexpr, which would make this compile-time or at least automatically const static

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    // create the colormap with 3 colors
+    ListColorMap *cmap = new (std::nothrow) ListColorMap();
+    cmap->init(3);
+    cmap->set(0, 0.0, 0, 0, 0, 255);
+    cmap->set(1, 0.004, 255, 255, 255, 255);
+    cmap->set(2, 1.0, 255, 255, 255, 255);
+
+    // create the image (logic representation)
+    IImage *im = new image();
```
DX-MON, line 79: This begs to be `auto im{std::make_unique<IImage>()};`

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
...
+    dmat4 rot_matrix = rotated_matrix(pos_params);
+    rot_matrix = rot_matrix / im->totalXres();
+    dvec4 deltax = rot_matrix[VX];
+    dvec4 deltay = rot_matrix[VY];
+    dvec4 delta_aa_x = deltax / 2.0;
+    dvec4 delta_aa_y = deltay / 2.0;
+    dvec4 topleft = center - deltax * im->totalXres() / 2.0 - deltay * im->totalYres() / 2.0;
+    topleft += delta_aa_x + delta_aa_y;
+
+    int w = im->Xres();
```
DX-MON, line 96: `const auto` is your friend.

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
...
+                pos.n,
+                100, // max iters
+                -1, // wrap param
+                min_period_iters, // min period iters (half the max)
+                1.0E-9, // period tolerance
+                x, y, 0, // x, y and aa: these values are not needed in the formula but required as arguments for debugging purposes
+                &iters_taken, &fate, &dist, &solid,
+                &direct_color, &colors[0]);
+            // process the formula output and get the value from colormap
+            rgba_t color;
```
DX-MON, line 129: Uninitialised. `rgba_t color{};` would guarantee a sane value and no UB.

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
...
+            im->setIndex(x, y, 0, dist);
+            // put the pixel color into the image buffer position
+            im->put(x, y, color);
+        }
+    }
+
+    // save the image
+    FILE *image_file = fopen("./output/mandelbrot.png", "wb");
+    image_file_t image_file_type = FILE_TYPE_PNG;
+    ImageWriter *image_writer = ImageWriter::create(image_file_type, image_file, im);
```
DX-MON, line 156: `std::unique_ptr<ImageWriter> image_writer{ImageWriter::create(...)};` would be more ideal with the automatic memory management this gives you.

examples/cpp/raw_mandelbrot.cpp
```diff
@@ -0,0 +1,175 @@
...
+    {
+        fprintf(stderr, "Error loading formula symbols: %s", dlerror());
+        dlclose(lib_handle);
+        return -1;
+    }
+    pf_obj *pf_handle = pfn();
+
+    // formula params: [0, 4.0, 0.0, 1.0, 4.0, 0.0, 1.0]
+    int param_len = 7;
+    struct s_param *params = (struct s_param *)malloc(param_len * sizeof(struct s_param));
```
DX-MON, line 46: `auto params{std::make_unique<s_param []>(param_len)};` for both automatic memory management, and non-UB behaviour. malloc() does not start object lifetimes, so every operation to params after this is undefined behaviour.
Only new/new [] can start an object's lifetime in C++.

examples/cpp/raw_mandelbrot.cpp
mindhells: You are right, it seems it was deprecated long ago but kept in order to allow compatibility. 
Also, cstdlib uses namespace so you can keep yours cleaner.
There are some uses of stdlib.h spread around fract4dc classes in the codebase, I will try to change those too.

examples/cpp/raw_mandelbrot.cpp
mindhells: IImage is an abstract but using smart pointers make sense 👍

examples/cpp/raw_mandelbrot.cpp
mindhells: If I make it constant then I had to cast it to non-const to use it in `pf_handle->vtbl->init` (with const_cast for example).
Would that make it any better?

DX-MON: ah, yeah.. avoid const_cast<>() - it's a bastion of instant undefined behaviour. I would still make it static so you don't pay for constructing it every call.

This problem suggests the vtbl interface needs redesigning for const-correctness as that stinks of bugs (a code-smell as lefticus would say)

examples/cpp/raw_mandelbrot.cpp
DX-MON: Ah, so it is.. in that case, `std::unique_ptr<IImage> im{std::make_unique<image>()};` is the preferred form here. This compiles and works as long as image is at some point derived from IImage, so this ensures convertability.

DX-MON:
My comments on malloc() usage still apply thanks to object lifetimes in the C++ VM, but your review of C compat head usage looks good otherwise

examples/cpp/raw_mandelbrot.cpp
mindhells: I'm working on an example that includes a custom formula code. That's something to have in mind.
To apply that improvement to gnofract we need to change the compiler, I think it shouldn't be very complicated (I guess that is part of a literal template of the C code generated for the formula).

mindhells:
I've applied most of the suggestions from @DX-MON to the examples (need to undo the const_cast though). 
I think it makes sense to apply them to the extension codebase (sources under fract4d/c ... I like to call it "gnofract engine") itself. That would take some time so I would leave that out of the scope of this PR.

examples/cpp/raw_mandelbrot.cpp
DX-MON: That works. And yeah, from what I recall about the code gen, it's simple enough to change this.

DX-MON:
Looks much better, thank you. the rbga_t comment remains, but otherwise approved for merge.

examples/cpp/custom_formula.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    // load formula lib
+    void *lib_handle = dlopen("./formula.so", RTLD_NOW);
+    if (!lib_handle)
+    {
+        fprintf(stderr, "Error loading formula: %s", dlerror());
+        return -1;
+    }
+    pf_obj *(*pfn)(void);
+    pfn = (pf_obj * (*)(void)) dlsym(lib_handle, "pf_new");
```
DX-MON, line 47: This wants to be a reinterpret_cast<>() - one of the few non-UB uses of this cast type. See point 8 of the explantation section of https://en.cppreference.com/w/cpp/language/reinterpret_cast

mindhells: There's a lot of C-style stuff like this casting in the fract4dc module. As far as I know, in this case, C-style cast will end up behaving like reinterpret_cast, but I fully agree with you reinterpret_cast should be used explicitly for the sake of clarity.

DX-MON: Unfortunately, the C-style cast actually invokes UB and issues a warning as it does a bit more than just a reinterpret_cast<>() - but yeah, this is something to look into and resolve.

I wouldn't want to put the C and C++ parts through a linter at the moment as it would be an overwhelming lot of output..

mindhells:
Hey @edyoung @DX-MON I don't have write access so, if you think the PR is ok to merge, please proceed

edyoung:
As far as I can tell you should have the same access as a member of the 'fract4d' org. I don't mind granting write access - thought you already had it.

PR #86, opened 2020-04-22
Add -std=c++17 building flag and update readme

guanchor:
There is no minimal C++ compiler version required and C++98 std is being used as default std in GCC versions prior to 6.1, so a new flag is added in the building phase to allow to support newer C++ language standards as used in #85.

Update the readme with info about pylint (more info #66).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -114,7 +114,7 @@ defines = [
 ]
 
 libraries = ['stdc++']
-extra_compile_args = ['-Wall', '-O0'] + png_flags
+extra_compile_args = ['-Wall', '-O0', '-std=c++17'] + png_flags
 define_macros = defines + extra_macros
 
 module_fract4dc = Extension(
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -114,7 +114,7 @@ def call_package_config(package, option, optional=False):
 ]
 
 libraries = ['stdc++']
-extra_compile_args = ['-Wall', '-O0'] + png_flags
+extra_compile_args = ['-Wall', '-O0', '-std=c++17'] + png_flags
```
edyoung, line 117: just noticed this. We definitely don't want -O0 in most cases.

mindhells: I think that's actually shadowed by distutils when creating the final compiler argument list. Still, I agree we don't want that there.

edyoung:
Generally speaking if a PR is approved feel free to merge at will.

PR #91, opened 2020-05-10
Fix #90

### LINKED ISSUE #90: Installed Gnofract 4D 4.1 fails to start - No such file or directory: 'fract4d/c/pf.h'
https://github.com/fract4d/gnofract4d/blob/b3b1bff199d4cce3a3bbe1d13ae7103e1ce7be3c/fract4d_compiler/codegen.py#L243-L244

codegen.py is installed by setup.py in:

"/usr/lib/python3.8/site-packages/fract4d_compiler/codegen.py"

and is trying to open files that aren't installed from a relative path that doesn't exist.

edyoung:
At runtime the C compiler needs the contents of pf.h and fract_stdlib.h. Originally I did this by including the contents of those files in here-strings in the python module. During refactor this was changed to load them from a subdir. But we weren't installing those files, and we weren't looking for them in the right place after installation, so that failed.

This installs the files as part of fract4d module and has the C compiler doing the includes rather than doing it ourselves.

### DIFF
--- a/fract4d_compiler/codegen.py
+++ b/fract4d_compiler/codegen.py
@@ -3,6 +3,7 @@
 # Generate C code from a linearized IR trace
 
 import re
+import os
 
 from . import absyn, fracttypes, ir, optimize
 from .fracttypes import Bool, Int, Float, Complex, Hyper, Color, IntArray, FloatArray, ComplexArray, VoidArray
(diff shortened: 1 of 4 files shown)


fract4d_compiler/codegen.py
```diff
@@ -3,6 +3,7 @@
 # Generate C code from a linearized IR trace
 
 import re
+import os
```
cjmayo, line 6: Don't think os is used.

PR #55, opened 2020-02-03
Remove unused code and extract new fract4d_compiler package

mindhells:
All files regarding unfinished and unmaintained gmp support has been deleted.
All compiler modules has been moved to a new package named fract4d_compiler alongside with their test files. Imports had been updated accordingly. 
Also the browser_module module has been moved to fract4dgui where it belongs.

PR #93, opened 2020-05-11
Add back lextab.py and parsetab.py

cjmayo:
These cannot be written when gnofract4d has been installed.

Partial revert of:
f9abe6d ("Remove unused code and extract new fract4d_compiler package (#55)", 2020-02-07)

WARNING: Couldn't write lextab module 'fract4d_compiler.lextab'. [Errno 13] Permission denied: '/usr/lib/python3.8/site-packages/fract4d_compiler/lextab.py'
WARNING: Couldn't open 'parser.out'. [Errno 13] Permission denied: '/usr/lib/python3.8/site-packages/fract4d_compiler/parser.out'
WARNING: Token 'COMMENT' defined, but not used
WARNING: Token 'ESCAPED_NL' defined, but not used
WARNING: Token 'COMMENT_FORMULA' defined, but not used
WARNING: There are 3 unused tokens
Generating LALR tables
WARNING: 13 shift/reduce conflicts
WARNING: 4 reduce/reduce conflicts
WARNING: reduce/reduce conflict in state 70 resolved using rule (parmlist -> parm)
WARNING: rejected rule (empty -> <empty>) in state 70
WARNING: Couldn't create 'fract4d_compiler.parsetab'. [Errno 13] Permission denied: '/usr/lib/python3.8/site-packages/fract4d_compiler/parsetab.py'

---

I don't understand what the merits are of generating these at run time, but that is not possible if you have installed with setup.py.

Is the .gitignore change wanted? Could be interesting if you have generated different files, or annoying if the differences aren't relevant.

Maybe there is another way, adding them only when building the tarball (ensuring they are present)?

I took them from gnofract4d-4.1.tar.gz.


cjmayo:
To clarify: the warning messages above are from installing a tarball I created myself from a clean repository using `setup.py sdist`.

The v4.1 release tarball (referred to above as gnofract4d-4.1.tar.gz) does include the files.

mindhells:
Damn, I always forget to test the "setup.py install" path. Could it be a possible to update travis config to use install instead build?
I like your idea @cjmayo of creating those files at installation time. It seems those are created because of the use of the [optimized mode](http://www.dabeaz.com/ply/ply.html#ply_nn15). So I understand they are created once and then reused.
You can trigger their creation:
```python
from fract4d_compiler import fractlexer, fractparser
fractlexer.get_lexer()
```
Does it makes sense to include their creation as a part of the setup?

cjmayo:
> Damn, I always forget to test the "setup.py install" path. Could it be a possible to update travis config to use install instead build?

I haven't tried but I think that would have to be an extra step because of running the tests - I don't think it would be right to run those against the installed gnofract4d.

> I like your idea @cjmayo of creating those files at installation time. It seems those are created because of the use of the [optimized mode](http://www.dabeaz.com/ply/ply.html#ply_nn15). So I understand they are created once and then reused.
> You can trigger their creation:
> 
> ```python
> from fract4d_compiler import fractlexer, fractparser
> fractlexer.get_lexer()
> ```
> 
> Does it makes sense to include their creation as a part of the setup?

In an ideal world this could be really clever but perhaps the simplest would be to just create a wrapper script to build the distribution (which would be easy to remove later).

Especially as this may all be going away:

https://github.com/dabeaz/ply/blob/83c5b91972a0905d620f3407e7d37f4cbcfe4178/CHANGES#L32-L39

Unfortunately it looks like there hasn't been an actual 4.0 release of PLY yet, so perhaps a risk to update to this code, although the changes don't sound too major.

There are a few ideas in [Packaging Advice](http://www.dabeaz.com/ply/ply.html#ply_nn49)

I don't think I will be doing any more on this myself, and also key here is how Ed wants to make releases.

edyoung:
Checking them in is simple and works. Let's do that for now.

PR #94, opened 2020-05-11
Include tests in the source distribution

cjmayo:
A distribution like Gentoo Linux wants to be able to run the tests themselves.

I removed the tests from the installation and missed this...


cjmayo:
Added testdata/formulas/test_bad_pp.frm, needed for fract4d_compiler/tests.

PR #97, opened 2020-05-11
Update fractutils for fract4d_compiler and testdata

cjmayo:
I don't use this but these are clearly out of date.

### DIFF
--- a/fractutils/parfile.py
+++ b/fractutils/parfile.py
@@ -9,7 +9,8 @@
 
 import math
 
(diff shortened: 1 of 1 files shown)


edyoung:
This code is probably never going to work. Parsing fractint .par files is a monumentally complicated task and the number of people who would be interested nowadays is pretty small. 

Be awesome to figure out how to parse UltraFractal files tho...

Thanks!

PR #99, opened 2020-05-12
Enable pytest coverage checking

cjmayo:
Test coverage is quite good, 79% with a lot of files over 90%, the overall number brought down by a few specific  files.

I suggest enabling reporting to keep it high.

I like the Coveralls service which can be hooked up to GitHub and provide browesable results and noitfications that can be used as PR checks https://coveralls.io/.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+[run]
+omit =
+	*/tests/*
+	fract4d_compiler/lex.py
+	fract4d_compiler/yacc.py
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.so
 *~
 #*
+.coverage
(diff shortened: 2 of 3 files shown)


DX-MON:
The other option would be [codecov.io](https://codecov.io/) which already plugs into Actions with a simple Bash script and has a really nice back-end interface

edyoung:
Thanks! open to any other integration anyone wants to experiment with.

mindhells:
Hey guys, one thought about the test coverage:
If I'm not wrong, many of the python tests are there for the fract4dc module. Meaning they are "covering" the C++ code under fract4d/c.
Have you ever though about creating C++ test for the fract4dc extension instead using python?
Which testing/mocking framework would you use in that case?

PR #108, opened 2020-05-16
Fix #107: wait until fd has something to read or fail

### LINKED ISSUE #107: Intermittent failing test: testFDSite
[This error](https://travis-ci.org/github/fract4d/gnofract4d/jobs/687139333#L567) appears from time to time. I've been looking for the conditions to reproduce it with no luck so far.

```bash

    ______________________________ Test.testFDSite ________________________________
    self = <fract4d.tests.test_fract4d.Test testMethod=testFDSite>
        def testFDSite(self):
            xsize = 64
            ysize = int(xsize * 3.0 / 4.0)
            im = image.T(xsize, ysize)
            (rfd, wfd) = os.pipe()
            site = fract4dc.fdsite_create(wfd)
        
            file = self.compileColorMandel()
        
            for x in range(2):
                handle = fract4dc.pf_load(file)
                pfunc = fract4dc.pf_create(handle)
                fract4dc.pf_init(pfunc, pos_params, self.color_mandel_params)
                cmap = fract4dc.cmap_create(
                    [(0.0, 0, 0, 0, 255),
                     (1 / 256.0, 255, 255, 255, 255),
                     (1.0, 255, 255, 255, 255)])
        
                fract4dc.calc(
                    params=[0.0, 0.0, 0.0, 0.0,
                            4.0,
                            0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
                    antialias=0,
                    maxiter=100,
                    yflip=0,
                    nthreads=1,
                    pfo=pfunc,
                    cmap=cmap,
                    auto_deepen=0,
                    periodicity=1,
                    render_type=0,
                    image=im._img,
                    site=site,
                    asynchronous=True)
        
                nrecved = 0
                while True:
                    if nrecved == x:
                        # print "hit message count"
                        fract4dc.interrupt(site)
        
                    nb = 2 * 4
                    bytes = os.read(rfd, nb)
                    if len(bytes) < nb:
                        self.fail(
                            "bad message with length %s, value %s" %
    >                       (len(bytes), bytes))
    E                   AssertionError: bad message with length 4, value b'\x03\x00\x00\x00'
    fract4d/tests/test_fract4d.py:579: AssertionError
```

In `gtkfractal.py` we have the `onData`callback (the file descriptor associated with the site is watched). 1st thing this method tries to read from the pipe, if it doesn't get the full message (this appears to be the problem in this test) it puts the data into a buffer class member and waits till the next call.

In this test we call `fract4dc.interrupt(site)` which turns a flag `interrupted=true` in the underlying C++ object. This doesn't stop the calculation job right away but instead lets the worker thread (see the calc function arguments contains `asynchronous=True`) to no proceed to the next "iteration"... so we have no guarantee the message has been written by the time we try to read it in the next instruction `bytes = os.read(rfd, nb)`.

mindhells:
see #107 
Note that I've removed the `fract4dc.interrupt(site)`since the test is reading messages util the "fractal is done" message is received.
Also we don't need to repeat pf_load and pf_create on every iteration.
Now we use the same strategy as in `gtkfractal.py` (buffering the message parts until we have the whole) plus we use `select` package to set a timeout in case C++ crashes to not to keep waiting forever.
While I was writing this message, I found the test failed because of the condition "no one on the other side", meaning C++ crashed somehow or I did something wrong :/
https://travis-ci.com/github/HyveInnovate/gnofract4d/jobs/335546177

### DIFF
--- a/fract4d/tests/test_fract4d.py
+++ b/fract4d/tests/test_fract4d.py
@@ -3,6 +3,7 @@
 import os.path
 import struct
 import math
+import select
 
 from . import testbase
 
@@ -539,9 +540,10 @@ class Test(testbase.ClassSetup):
 
         file = self.compileColorMandel()
 
+        handle = fract4dc.pf_load(file)
+        pfunc = fract4dc.pf_create(handle)
+
         for x in range(2):
-            handle = fract4dc.pf_load(file)
-            pfunc = fract4dc.pf_create(handle)
             fract4dc.pf_init(pfunc, pos_params, self.color_mandel_params)
             cmap = fract4dc.cmap_create(
                 [(0.0, 0, 0, 0, 255),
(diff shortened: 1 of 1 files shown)


mindhells:
Btw I used this to reproduce the error in 1st place:
```bash
for i in {1..1000}; do pytest fract4d/tests/test_fract4d.py::Test::testFDSite -s; done 
```
I found an error rate of 1~2% in my environment.

DX-MON:
Honestly hadn't thought to use select() or another polling mechanism.. smart choice!

Issue #78, opened 2020-04-11
GPU support


edyoung:
It could be possible to generate code for Nvidia and/or AMD GPUs. Since fractals are highly parallelizable, this could be substantially faster. It would be quite a lot of work. OpenCL and CUDA are possibilities for the backend.

PR #112, opened 2020-05-23
Starting point for a benchmark to compare multiple-precision libs

edyoung:
Uses Google benchmark library. Need to rearrange the docker stuff, it recompiles the benchmark library every time which slows things down a lot.

Run the benchmark with `examples/cpp/mp_benchmark.sh`

Sample output:
```
Running ./example
Run on (4 X 2700 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x2)
  L1 Instruction 32 KiB (x2)
  L2 Unified 256 KiB (x2)
  L3 Unified 3072 KiB (x1)
Load Average: 2.38, 1.75, 1.11
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
***WARNING*** Library was built as DEBUG. Timings may be affected.
----------------------------------------------------------------------
Benchmark                            Time             CPU   Iterations
----------------------------------------------------------------------
BM_fractal/min_time:120.000       6823 ms         6823 ms           25
```

### DIFF
--- a/doc/mathlibs.md
+++ b/doc/mathlibs.md
@@ -1,0 +1,34 @@
+
+# Notes on Math Libraries
+
+I went through the list of all arbitrary precision math libs at https://en.wikipedia.org/wiki/List_of_arbitrary-precision_arithmetic_software to see which might work best for Gnofract 4D.
+
+**Criteria**:
+- Fast
+- Supports floating point
+- Target numbers up to 4000 bits (?)
+- Need trig and other elementary functions
+
+
+**Library** | **Suitable** | **Notes**
+--- | --- | ---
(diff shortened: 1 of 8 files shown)


mindhells:
bringing [Boost](https://www.boost.org/) into the project opens a world of possibilities, like:
https://github.com/boostorg/compute (in regard to #78)

DX-MON:
I will warn that it also severely complicates things - particularly when trying to get the project to build on a RHEL or SLES box because they invariably have very old Boost's and not installed by default even when you put the compilers on..

But as long as that's a trade-off you're willing to make, it is an intriguing idea

PR #118, opened 2020-06-01
Run pylint as a separate job

cjmayo:
Report linting and testing results separately. Only run pylint once.


cjmayo:
Update to use Python 3.7 which is included in the Travis image to save a bit of time.

PR #122, opened 2020-06-06
Open an About Dialog from Help About

cjmayo:
Help About currently launches the HTML manual (to the About section) as
does Help Contents. Instead have it open a standard About Dialog.

BSD 3 license type not currently accesible in GTK 3.24.0:
https://gitlab.gnome.org/GNOME/gtk/-/issues/2734

---

I tried to use existing sources to populate the dialog - do suggest changes, especially the description, a challenge to come up with something short. Here's how it looks:

Gnofract4D Help About


edyoung:
looks good to me, thanks

Issue #126, opened 2020-06-08
MacOS: pixmaps not loading


mindhells:
Pixmaps are not being loaded in MacOS environment (10.15.3).

![Captura de pantalla 2020-06-08 a las 10 47 57](https://user-images.githubusercontent.com/1893197/84016818-bfa8cd00-a975-11ea-83d4-93405e2fcf89.png)

![Captura de pantalla 2020-06-08 a las 10 48 27](https://user-images.githubusercontent.com/1893197/84016831-c2a3bd80-a975-11ea-9870-cf4a233f6ab3.png)

cjmayo:
Summary:
The application pixmaps are loaded from share/gnofract4d/pixmaps.
The logo on the About dialog is loaded by GTK from the desktop environment theme (the advantages there being it also supports the icon displayed by the environment in e.g. task switcher, application list or menus and manages different sized icons).

Info from #127:
The share folder is under /usr/local/lib/python3.7/site-packages/gnofract4d-4.2-py3.7-macosx-10.15-x86_64.egg. This share folder contains the manual and also the pixmaps

I don't know if this is useful, but where does MacOS place the application pixmaps if installed with:
`setup.py install --single-version-externally-managed`

cjmayo:
Possibly even:
`setup.py install --single-version-externally-managed --root=/`

https://stackoverflow.com/a/33791008

mindhells:
It's interesting because when I do that no icon are drawn in the user interface (it also fails to find color maps or any other data file). While doing a simple install only a few icons are missing.

**with `setup.py install --single-version-externally-managed --root=/`**

```shell
...
copying pixmaps/logo/48x48/gnofract4d.png -> /usr/local/share/pixmaps
copying pixmaps/logo/16x16/gnofract4d.png -> /usr/local/share/icons/hicolor/16x16/apps
copying pixmaps/logo/32x32/gnofract4d.png -> /usr/local/share/icons/hicolor/32x32/apps
copying pixmaps/logo/48x48/gnofract4d.png -> /usr/local/share/icons/hicolor/48x48/apps
creating /usr/local/share/icons/hicolor/64x64
creating /usr/local/share/icons/hicolor/64x64/apps
copying pixmaps/logo/64x64/gnofract4d.png -> /usr/local/share/icons/hicolor/64x64/apps
creating /usr/local/share/icons/hicolor/128x128
creating /usr/local/share/icons/hicolor/128x128/apps
[...]

PR #127, opened 2020-06-08
Fix relative manual installation directory

cjmayo:
Installed into the system in e.g.:

/usr/lib/python3.8/site-packages/fract4dgui/

----

I thought this was working - not sure what changed here. Is this OK for everyone?

(manual looks very modern now! Definitely the right decision to bundle the HTML, I did install Hugo to try it myself but a lot of Go modules which would not make a great dependency for Gnofract 4d installation on Gentoo).

### DIFF
--- a/fract4dgui/main_window.py
+++ b/fract4dgui/main_window.py
@@ -1274,7 +1274,7 @@ class MainWindow:
 
         # look locally first to support run-before-install
         local_dir = "manual/public/"
-        install_dir = "../../share/gnome/help/gnofract4d/C/"
+        install_dir = "../../../../../share/gnome/help/gnofract4d/C/"
 
         helpfile = fractconfig.T.find_resource(
             base_help_file, local_dir, install_dir)


cjmayo:
Another way to go:
```py
install_dir = os.path.join(sys.base_prefix, "share/gnome/help/gnofract4d/C/")
```

cjmayo:
> Another way to go:

Probably best to ignore my ramblings here for now and fix as is.

mindhells:
This change wouldn't work on MacOS.

Path before change:
`file:///usr/local/lib/python3.7/site-packages/gnofract4d-4.2-py3.7-macosx-10.15-x86_64.egg/share/gnome/help/gnofract4d/C/index.html`

Path after change:
`file:///usr/local/lib/share/gnome/help/gnofract4d/C/index.html`

The share folder is actually under `/usr/local/lib/python3.7/site-packages/gnofract4d-4.2-py3.7-macosx-10.15-x86_64.egg`. This share folder contains the manual and also the pixmaps which, I don't know why, are not found #126.

cjmayo:
OK, that's a no to that. I think there are a few options, I'll try and post some more soon.

cjmayo:
This is just fixing the symptom and not the root cause.

PR #128, opened 2020-06-08
Use run() with aboutDialog

cjmayo:
@mindhells does that work on MacOS?

I spent too much time looking at the content of the dialog and didn't spot I had used show().


mindhells:
It works just fine

cjmayo:
Thanks for testing.

PR #101, opened 2020-05-13
Change documentation to markdown

edyoung:
We now have a directory 'manual/content' containing the docs in markdown format. Setup.py uses hugo to generate html docs in manual/public. The generated doc and associated files are installed when you run setup.py installed. 

TODO: 
- update website

PR #129, opened 2020-06-09
Ignore autogen files

mindhells:
Seems that the whole manual/public folder can be ignored.

Also some egg files where introduced in this #101 and could be ignored as well.


mindhells:
I think I've just realized the reason behind having a version of the manual already generated.

Setup needs those files and adding the createdocs script into the setup process add a new "must have system dependency": hugo.

PR #130, opened 2020-06-09
Install manual in share/gnofract4d/help

cjmayo:
Using a web browser to view now, not Yelp. Make the path consistent with
other Gnofract 4D data to allow simplification.

--- 

This doesn't fix the problem (could temporarily add the dots for Linux?) but it is an enabling step to improving the code that finds data.


cjmayo:
And a bonus, favicon.png was executable.

Issue #119, opened 2020-06-04
C++ house naming convention


mindhells:
There're different naming conventions in use in the C++ layer of Gnofract4D.
Some examples on user-defined types:
- MTFractWorker, IImage ...
- s_pixel_stat, s_rgba, s_pf_data ...
- fractFunc, pointFunc ...
- image_reader, image_writer ...

on members:
- nThreads
- m_totalXres, m_totalYres
- auto_deepen
- lookup_with_transfer(),
- totalXres(), getBuffer(),

edyoung:
As you've noticed, the naming style is quite inconsistent today, I guess because I have never been that religious about it. In general I would be happy to see more or less any consistent coding style adopted.  

Ideally:
- We adopt one off the shelf rather than defining one
- it would be one where there was some tooling available to point out inconsistencies
- any changes to update areas of code to conform to style would _only_ do that, not mixed in with other changes, even refactoring - makes code review much easier.

https://clang.llvm.org/extra/clang-tidy/ might be worth a look

PR #134, opened 2020-06-12
Fix unused variable and extra ‘;’ warnings

cjmayo:
Please check - I'm just following the compiler messages.


DX-MON:
Changes look good - the warnings should indeed be headed.. I'd suggest enabling `-Wall -Wextra -pedantic` and see what falls out as I expect those aren't the only warnings that need looking at

mindhells:
> Changes look good - the warnings should indeed be heeded.. I'd suggest enabling `-Wall -Wextra -pedantic` and see what falls out as I expect those aren't the only warnings that need looking at

I'm planning to try clang-tidy as @edyoung suggested in #119 . That should help.

PR #135, opened 2020-06-13
remove fract4d_stdlib

edyoung:
Complicated story: at runtime, we load a generated shared
object containing the code for the fractal. We want that to
be able to call some helper functions that are contained in
the fract4dc module. That module is also dynamically loaded,
and by default Python loads it without specifying RTLD_GLOBAL.
That means functions in it are not visible to other dynamically
loaded modules, which we need.

Originally I resolved this with a terrible hack: create another .so
containing the code I wanted, and dlopen() that specifically
with RTLD_GLOBAL. Looks like there is an easier fix, using
sys.setdlopenflags, which I never realized. Thanks @mindhells !

This simplifies things by removing that extra complication. Sweet.

### DIFF
--- a/fract4d/__init__.py
+++ b/fract4d/__init__.py
@@ -1,3 +1,8 @@
 #!/usr/bin/env python3
 
 # package file for fract4d
+
+import sys
+import os
+
+sys.setdlopenflags(os.RTLD_GLOBAL | os.RTLD_NOW)
--- a/fract4d/c/fract4dmodule.cpp
+++ b/fract4d/c/fract4dmodule.cpp
@@ -13,9 +13,6 @@
 
 #include "Python.h"
 
-#include "cmap_name.h"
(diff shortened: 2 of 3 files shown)


fract4d/__init__.py
```diff
@@ -1,3 +1,8 @@
 #!/usr/bin/env python3
 
 # package file for fract4d
+
+import sys
+import os
+
+sys.setdlopenflags(os.RTLD_GLOBAL | os.RTLD_NOW)
```
cjmayo, line 8: Missing final newline.

cjmayo:
Has Travis dropped off again? 

Tests pass locally for me for this one. There may be an issue with pylint (although that could be my setup).

```
ImportError: Unable to find module for fract4d/fract4dc.cpython-37m-x86_64-linux-gnu.so in gnofract4d,
```

PR #136, opened 2020-06-14
Add gradient options

edyoung:
This change does the following:

1) Adds some .ase palette files I found pleasing
2) Adds the option to edit a gradient online in https://coolors.co

The UX for this is not super discoverable, and could do with some work. But their gradient editor is better than the janky one in Gnofract 4D, so I found this useful.


fract4d/gradient.py
```diff
@@ -523,6 +561,7 @@ def load_list(self, l, maxdiff=0):
                 if self.compare_colors(
                         test_segment.get_color_at(last_index), last_color, maxdiff):
                     # can compress, update in place
+                    #print("compress!")
```
cjmayo, line 564: Debug code left in?

fract4dgui/main_window.py
```diff
@@ -1265,7 +1278,6 @@ def command_reference(self, *args):
     def report_bug(self, *args):
         url = "https://github.com/fract4d/gnofract4d/issues"
         utils.launch_browser(
-            self.userPrefs,
```
cjmayo, line 1268: Useful tidy up but really a separate change.

pylintrc
```diff
@@ -2,7 +2,6 @@
 disable=all
 enable=E,
   bad-whitespace,
-  trailing-whitespace,
```
cjmayo, line 5: Personal choice I guess.

fract4dgui/main_window.py
```diff
@@ -1208,6 +1210,17 @@ def redo(self, *args):
         """Redo an operation after undoing it."""
         self.model.redo()
 
+    def paste(self, *args):
+        """Paste (can be used to update colors)."""
+
+        clipboard = Gtk.Clipboard.get(Gdk.SELECTION_CLIPBOARD)
+        text = clipboard.wait_for_text()
```
cjmayo, line 1217: Returns

a newly-allocated UTF-8 string which must be freed with g_free(), or NULL if retrieving the selection data failed. (This could happen for various reasons, in particular if the clipboard was empty or if the contents of the clipboard could not be converted into text form.). 

Suggest adding a test for `is None`. Else if pasting non-text:

```
gnofract4d/fract4d/gradient.py", line 516, in load_from_url
    if not url.startswith("https://coolors.co/"):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

fract4d/gradient.py
```diff
@@ -495,6 +495,44 @@ def load_map_file(self, mapfile, maxdiff=0):
...
+        return "https://coolors.co/" + ("-".join(clist))
+
+    def load_from_url(self, url):
+        'Parse a URL which encodes a palette. We don\'t actually fetch it, the data is in the URL itself'
+        try:
+            if not url.startswith("https://coolors.co/"):
+                return False
+            url = url[19:] # remove domain
+            parts = url.split('-')
+            print(parts)
```
cjmayo, line 520: This does look like active debug code.

fract4d/gradient.py
```diff
@@ -495,6 +495,44 @@ def load_map_file(self, mapfile, maxdiff=0):
...
+            colorlist = []
+            ncolors = len(parts)
+            i = 0
+            for p in parts:
+                (r, g, b) = (int(p[0:2], 16), int(p[2:4], 16), int(p[4:6], 16))
+                entry = (i/ncolors, r, g, b, 255)
+                print(entry)
```
cjmayo, line 527: And this.

PR #141, opened 2020-06-21
Improve dependency documentation

cjmayo:
More comments in the code, fixes and tidying in README.md and some better run-time messages.

A bit more explanation in some of the commit messages.

PR #142, opened 2020-06-22
Use GitHub Actions instead of Travis

cjmayo:
With all the problems with Travis here is an attempt to use GitHub actions instead. Let's see if the PR works. One advantage - it is easier to re-run when there is a segfault...

It does clash with #141. So, would need to be updated.

### DIFF
--- a/.github/workflows/test-gnofract4d.yml
+++ b/.github/workflows/test-gnofract4d.yml
@@ -11,15 +11,16 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest]
-        python: [3.5, 3.6, 3.7]
+        os: [macos-10.15, ubuntu-20.04]
+        python: [3.7]
         toxenv: [py]
-        exclude:
-          - os: macos-latest
-            python: 3.5
-          - os: macos-latest
-            python: 3.6
         include:
+          - os: ubuntu-18.04
+            python: 3.6
+            toxenv: py
+          - os: ubuntu-16.04
+            python: 3.5
+            toxenv: py
           - os: ubuntu-latest
             python: 3.7
             toxenv: pylint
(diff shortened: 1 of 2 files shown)


mindhells:
Never used them but looks great.
Maybe we could add the OS into the matrix and include macos as a supporting OS as well as ubuntu. I would rather use specific versions though: `ubuntu-18.04` and `macos-10.15`
Artifacts sound promising too for building binary releases with pyoxidizer (as @edyoung suggested in #133).

cjmayo:
Added upload to Codecov (it can take the XML report from pytest-cov, unlike Coveralls).

Added a MacOS job for Python 3.7.

Being specific about OS versions would be my first idea, but here I'm not so sure, it creates maintenance work. On the other hand if we are saying we are supporting specific OS versions then that would be useful.

DX-MON:
You can matrix on the OS versions as well as Python and have -latest and specific versions so you get to see if anything broke in the newest of the OS and set up support guarantees for the rest

edyoung:
I'm all for it

cjmayo:
Rebased on master to fix clashes and changed to specific OS's with the objective of trying everything at least once (will Ubuntu 20.04 work???).

cjmayo:
> I'm all for it

Great! Can you enable Actions for the repository under Settings/Actions?

edyoung:
Already enabled as as far as I can see

image

cjmayo:
Thanks for checking. I have now pushed this PR as a branch to the repository and the actions started running. Also realised another commit was needed to stop trying to upload to Codecov from the Pylint job.

edyoung:
Thanks! Looks great

mindhells:
should .travis.yml file be removed?

DX-MON:
Seeing we can re-create the config if we want to try fixing and re-enabling Travis at some later date, I see no harm in removing the file

PR #143, opened 2020-07-03
update min python to 3.6 

edyoung:
This allows use of f-strings for interpolation
Updated one file (absyn.py) to see how that looks
Updated readme and test control files to specify new min


edyoung:
Curious to see an apparently random segfault in the test runs. We must have an intermittent issue in there somewhere. Why does everything run on push and pull_request, is that useful? Guess it doesn;t hurt

cjmayo:
> Curious to see an apparently random segfault in the test runs. We must have an intermittent issue in there somewhere. Why does everything run on push and pull_request, is that useful? Guess it doesn;t hurt

Actions give you a lot of control, which also means there is plenty to figure out! It appeared that you need the pull_request one to get a check when the PR is created, and the push both for updates to master and updates to a PR - but maybe updates to a PR can be done by the synchronise event (more investigation needed).

.github/workflows/test-gnofract4d.yml
```diff
@@ -18,9 +18,6 @@ jobs:
           - os: ubuntu-18.04
             python: 3.6
             toxenv: py
-          - os: ubuntu-16.04
```
cjmayo, line 21: Yes, 3.5 is the standard Python 3 for Ubuntu 16.04 so this does mean dropping it.

fract4d_compiler/absyn.py
```diff
@@ -4,7 +4,6 @@
 
 from . import fracttypes
 
-
```
cjmayo, line 7: Two lines here and before top-level functions are as per PEP 8:
https://www.python.org/dev/peps/pep-0008/#blank-lines

fract4d_compiler/absyn.py
```diff
@@ -118,35 +105,30 @@ def __next__(self):
 
         return ret
 
-
```
cjmayo, line 121: PEP 8 line as above.

fract4d_compiler/absyn.py
```diff
@@ -118,35 +105,30 @@ def __next__(self):
...
         if not isinstance(tree.children, list):
-            raise Exception("children not a list: %s instead" % tree.children)
+            raise Exception(f"children not a list: {tree.children} instead")
         for child in tree.children:
             CheckTree(child, 0)
     return 1
 
 # shorthand named ctors for specific node types
 
-
```
cjmayo, line 136: Run of PEP 8 lines as above.

tox.ini
```diff
@@ -14,9 +15,10 @@ commands =
 
 passenv = DISPLAY XAUTHORITY HOME
 
+# imported by tests
```
cjmayo, line 18: It was specific to pytest so maybe reword to "pytest imported by tests"

PR #145, opened 2020-07-07
Remove dependency on xdg-open for launching browser

cjmayo:
I found this by running on a machine that didn't have xdg-open installed. With this change it could start a web browser.

The change removes the Mac specific code, @mindhells if you can check it still works, thanks.

### DIFF
--- a/.github/workflows/test-gnofract4d.yml
+++ b/.github/workflows/test-gnofract4d.yml
@@ -35,7 +35,8 @@ jobs:
       - name: Install Ubuntu packages
         if: ${{ runner.os == 'Linux' }}
         # libgirepository1.0-dev required to build PyGObject from source
-        run: sudo apt install gir1.2-gtk-3.0 libgirepository1.0-dev xvfb
+        # gvfs is needed for Gio.AppInfo.launch_default_for_uri()
+        run: sudo apt install gir1.2-gtk-3.0 gvfs libgirepository1.0-dev xvfb
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
--- a/fract4dgui/utils.py
+++ b/fract4dgui/utils.py
@@ -9,12 +9,11 @@
 
 import os
 import inspect
-import sys
 
 import gi
 gi.require_version('Gdk', '3.0')
 gi.require_version('Gtk', '3.0')
-from gi.repository import Gtk, Gdk, GLib
+from gi.repository import Gtk, Gdk, Gio, GLib
 
 from . import hig
 
@@ -161,17 +160,13 @@ def color256FromFloat(r, g, b, color):
 
 
 def launch_browser(url, window):
-    if sys.platform[:6] == "darwin":
-        cmd = 'open "%s" &' % url
-    else:
-        cmd = 'xdg-open "%s" &' % url
     try:
-        os.system(cmd)
-    except Exception as err:
+        Gio.AppInfo.launch_default_for_uri(url)
+    except GLib.Error as err:
         d = hig.ErrorAlert(
             primary=_("Error launching browser"),
-            secondary=_("Try copying the URL '%s' manually to a browser window.\n" % url) +
-            str(err),
+            secondary=_(f"Try copying the URL '{url}' manually to a browser window.\n"
+                        f"{err.message}"),
             transient_for=window)
         d.run()
         d.destroy()


cjmayo:
This is repeatedly causing the Ubuntu jobs to timeout. The only real change I can see is the `import Gio` - all the dependencies I can think of are installed. Any ideas?
I'm wary of merging it and then breaking any future checks.

mindhells:
try `docker/bin/run_tests.sh`
That script runs the tests within a docker container based on an Ubuntu 18.04 image.

I can tell you I have the same problem as in the pipeline, it gets stuck here:
```bash
fract4d/fract4d_new/tests/test_fract4dc.py ..                                                                                                                               [  0%]
fract4d/fract4d_new/tests/test_fract4dc_controller.py ...                                                                                                                   [  0%]
fract4d/fract4d_new/tests/test_image_wrapper.py .                                                                                                                           [  1%]
fract4d/tests/test_3d.py ....                                                                                                                                               [  1%]
fract4d/tests/test_animation.py ..........                                                                                                                                  [  3%]
fract4d/tests/test_colorizer.py ...                                                                                                                                         [  3%]
fract4d/tests/test_event.py ....                                                                                                                                            [  4%]
fract4d/tests/test_formsettings.py ...................                                                                                                                      [  7%]
fract4d/tests/test_fract4d.py ..............s................                                                                                                               [ 12%]
fract4d/tests/test_fractal.py ..........................................................                                                                                    [ 22%]
fract4d/tests/test_fractconfig.py .........                                                                                                                                 [ 24%]
fract4d/tests/test_gradient.py ....................................                                                                                                         [ 30%]
fract4d/tests/test_image.py ..................                                                                                                                              [ 33%]
fract4d/tests/test_options.py ........                                                                                                                                      [ 34%]
fract4dgui/tests/test_angle.py ....                                                                                                                                         [ 35%]
fract4dgui/tests/test_browser.py .....                                                                                                                                      [ 36%]
fract4dgui/tests/test_browser_model.py ...............                                                                                                                      [ 38%]
fract4dgui/tests/test_director.py ....                                                                                                                                      [ 39%]
fract4dgui/tests/test_fourway.py ..                                                                                                                                         [ 39%]
fract4dgui/tests/test_gtkfractal.py ...............                                                                                                                         [ 42%]
fract4dgui/tests/test_hig.py .......                                                                                                                                        [ 43%]
fract4dgui/tests/test_main_window.py ..
```

nothing comes to my mind that can cause that, but if you figure out something you can try that script before pushing to github

cjmayo:
Thanks for pointer out the Docker tests, I hadn't tried those before.

It wasn't dependencies that was causing the problem the hang. I fixed it for Ubuntu in a container switching to the asynchronous `launch_default_for_uri_async`, but that now seems to break macOS in a container...

Actions haven't run here, I 'll try switching it from draft and back.

cjmayo:
Finally found the culprit, no GVfs. Easy to add that for Ubuntu but still a problem for macOS, couldn't find it in Hombrew so guessed nautilus would include it, although that is only available as a cask. That seemed to be working and now not...
@mindhells can you see what was providing GVfs for you?

mindhells:
> Finally found the culprit, no GVfs. Easy to add that for Ubuntu but still a problem for macOS, couldn't find it in Hombrew so guessed nautilus would include it, although that is only available as a cask. That seemed to be working and now not...
> @mindhells can you see what was providing GVfs for you?

I've tried this last change locally and I get "operation not supported by the current backend" error, so I guess no GVfs as well.
I think the nautilus cask you are installing (https://formulae.brew.sh/cask/nautilus) has actually nothing todo with GVfs but instead is a charting tool for docker compose.

Your 1st approach was working on MacOS so I would rather go using it when system is "darwin". You don't get a single solution for both systems but at least yo get rid of xdg-open dependency.

cjmayo:
> I've tried this last change locally and I get "operation not supported by the current backend" error, so I guess no GVfs as well.
> I think the nautilus cask you are installing (https://formulae.brew.sh/cask/nautilus) has actually nothing todo with GVfs but instead is a charting tool for docker compose.

Oops, thanks!
 
> Update:
> Actually you can [install nautilus file manager with Mac ports ](https://ports.macports.org/port/nautilus/summary), but that introduces a new dependency for MacOS

Interesting, but yes probably too much hassle just for this.

> Your 1st approach was working on MacOS so I would rather go using it when system is "darwin". You don't get a single solution for both systems but at least yo get rid of xdg-open dependency.

I've gone back to the 1st approach for everyone, adding gvfs for Ubuntu gets the tests passing on GitHub.

I think that will do for this PR. I have go more coming for Docker.

cjmayo:
The issue with Docker is no web browser!

PR #146, opened 2020-07-07
Recommend pip for installation instead of setup.py

cjmayo:
pip ensures the data files are installed in the expected locations and
not e.g. in an egg. It is also easier to use

---

Don't suggest sudo with pip

It is not a requirement and should be a user decision.

---

Based on the experience with issue #126. Plus a new one on sudo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,11 @@ You can then run Gnofract 4D in the local directory:
 
     ./gnofract4d
 
-You can also install it:
+Alternatively to build and install Gnofract 4D (version 4.2 as an example) type:
 
-    sudo ./setup.py install
+    pip3 install gnofract4d-4.2.tar.gz
 
-You can then run it as:
+You can then run it by clicking on the desktop icon or typing:
 
     gnofract4d
 
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Testing
 
 Testing requires pytest for Python 3. In some distributions, 'pytest' is for Python 2. Run
 
-    sudo pip3 install pytest
+    pip3 install pytest
 
 To get the latest.
 
@@ -84,7 +84,7 @@ You can run it in two ways:
 
  - Directly (firstly you should install pylint through pip3):
 
-    $ sudo pip3 install pylint
+    $ pip3 install pylint
     $ ./bin/pylint.sh
 
  - Using docker (same environment as pipeline):


README.md
```diff
@@ -12,17 +12,9 @@ Examples, screenshots etc are at https://fract4d.github.io/gnofract4d
...
-    ./setup.py build
-
-You can then run Gnofract 4D in the local directory:
-
-    ./gnofract4d
-
-You can also install it:
-
-    sudo ./setup.py install
+    pip3 install gnofract4d-4.2.tar.gz
```
mindhells, line 17: Doesn't work for me like that, it does instead if I do `pip3 install .` from the root folder. 
I think @DX-MON also mentioned that you can put there the URL of the repository there in #126 . That could also be worth mentioning.

On the other hand I would not remove the build + run instructions `./setup.py build` + `./gnofract4d`.

cjmayo: > Doesn't work for me like that, it does instead if I do `pip3 install .` from the root folder.

Might be worth investigating. I just tried it again on Ubuntu 20.04 and was OK.
The hope was it avoided the user having to do the unpacking.

> I think @DX-MON also mentioned that you can put there the URL of the repository there in #126 . That could also be worth mentioning.

It was me - there as an example of the power of pip, here was trying to avoid suggesting the user install things straight off the Internet.

> On the other hand I would not remove the build + run instructions `./setup.py build` + `./gnofract4d`.

I thought that might be controversial. To me that is a developer convenience that it would be easier not to have to support, it doesn't get all the icons etc. installed properly (I also had in mind a future solution like in the Meson PR with a module created at installation time with the absolute resource paths in and the version). But of course whatever the consensus is. There is a linting section later in the document, maybe it could be in an expanded developer section there.

mindhells: Maybe I'm watching this from 2 different perspectives: 

I would expect "normal" users to land in the public site: https://fract4d.github.io/gnofract4d/
There's a download button there but no instructions about how to install it, I guess.
I might be interesting to place there this [suggestion of yours:](https://github.com/fract4d/gnofract4d/issues/126#issuecomment-653788236) `pip3 install https://github.com/fract4d/gnofract4d/releases/download/v4.2/gnofract4d-4.2.tar.gz` or download the file and then `pip3 install gnofract4d-4.2.tar.gz`.

On the other hand, I see the Readme as a good place to provide developers with basic instructions on how to build and run the application locally.
Not sure if possible, but since [you can do this with pip](https://github.com/fract4d/gnofract4d/issues/126#issuecomment-653802580), maybe you can directly install a PR version to try something out.

Another approach is, as you said, to create an expanded developer section with that information. Actually, we have the [wiki](https://github.com/fract4d/gnofract4d/wiki) that could need some review.

I think this is an interesting discussion you brought out here.

edyoung: Possibly the single hardest thing about writing software is packaging the damn thing up and getting it to users :-(

We have several different cases

1. Individual user finds the github,io web page and wants to install. 

2. User finds the README on GitHub.com and wants to install

3. User downloads the .tar.gz and reads the README inside it

4. User clones the repo and reads the README looking for that info.

Users 1-3 just want to draw fractals as fast as possible. User 4 may want to contribute to the program.

To try and address these:

I think it makes sense to modify the front page 'Download' to provide more info rather than just linking directly to the release. This could be a link to the wiki page, which can include quick options like 'pip3 from URL'.  I can change that.

The README has to serve the other classes of user. I am OK with recommending pip3 over setuptools, but I personally would not remove the instructions for running it locally without installation, because that's the most convenient way IMHO. Also `pip3 install` without sudo puts the program in a dir which may not be on your $PATH and doesn't tell you which dir it is, which I found confusing. If there's any way to make that clearer it would help.

Any thoughts?

cjmayo: There certainly is no one easy solution for everyone.

> Possibly the single hardest thing about writing software is packaging the damn thing up and getting it to users :-(

With documentation being the second hardest...

> The README has to serve the other classes of user. I am OK with recommending pip3 over setuptools, but I personally would not remove the instructions for running it locally without installation, because that's the most convenient way IMHO. Also `pip3 install` without sudo puts the program in a dir which may not be on your $PATH and doesn't tell you which dir it is, which I found confusing. If there's any way to make that clearer it would help.

Yes, can add something about PATH if it doesn't work. Recent Debian and Ubuntu's do include `~/.local/bin` in PATH, so hopefully this will be a diminishing issue.

mindhells: > We have several different cases

I would say, there're 2 different kind of users:
- normal: users that want to install and use the program only
- advanced: users also interested in contributing or learning about the internals 

On the other hand there're 4 different sources of information at this moment:
- frontpage
- manual
- readme
- wiki

The frontend and the readme already reference each other, so either the user finds one or another he can get to the right spot easily. And the manual is part of both the frontend and the application itself.

> I think it makes sense to modify the front page 'Download' to provide more info rather than just linking directly to the release. This could be a link to the wiki page, which can include quick options like 'pip3 from URL'. I can change that.

I think the wiki is mostly outdated, though it holds some useful pieces of information. I would move those that still apply to the manual or the readme, depending if they are meant for the normal or the advanced user. Then I would drop the wiki and put the `pip install URL` in the frontpage. 

> Possibly the single hardest thing about writing software is packaging the damn thing up and getting it to users :-(

GIMP uses [flatpak](https://flatpak.org/) and [flink](http://www.finkproject.org/) for creating distributable packages for Linux and MacOS. No idea how hard could it be to put this project into these, does it make sense to create an issue and discuss about that separately?

cjmayo:
I have put the current `./setup.py build` and `./gnofract4d` back in the same place and just replaced `./setup.py install` with `pip3 install gnofract4d-4.2.tar.gz` - that's the bit I really want to fix, because of the the problems we've established.

I've checked and Fedora/RHEL have also had `./local/bin` in PATH for a while, it is present on CentOS 7. It is also part of the systemd docs: https://www.freedesktop.org/software/systemd/man/file-hierarchy.html#Home%20Directory.

PR #147, opened 2020-07-08
Fract4dc cleanup

mindhells:
So I started this to clean up some unneeded code in the workers, but then I got enthused and ended up adding some extra cleanings and tidying up:
- removed `_REENTRANT` define (see https://stackoverflow.com/questions/2601753/what-is-the-reentrant-flag)
> So it seems like _REENTRANT is mostly a no-op, these days. It might once have declared a lot of new functions, such as strtok_r; but these days those functions are mostly mandated by various decades-old standards (C99, POSIX 95, POSIX.1-2001, etc.) and so they're just always enabled.
- conditionally compile the experimental code in `STFractWorker`
- conditionally compile the `pixel_changed` callback interface for `Site` (this is meant for debug purposes only)
- narrow down the public interface of the `IFractWorker`
- added some extra warnings for compiling the **fract4dc** module: those helped me out to detect code smells tackled in this PR
- silenced some of the above warnings fixing them up or with C++ attributes
- drop support for **python 3.5** in the docker scripts
- by the way `setup.py` has been updated with f-strings 
- `setup.py `now uses `pkg_config` to link `libjpeg`
- added a debug mode to the `setup.py`: now **-O0** is no longer the default optimization level

Sorry to mix things up, hope you can have a look with no much trouble.

### DIFF
--- a/fract4d/c/fract4dc/arenas.cpp
+++ b/fract4d/c/fract4dc/arenas.cpp
@@ -7,7 +7,7 @@
 
 namespace arenas {
 
-    PyObject * pyarena_create(PyObject *self, PyObject *args)
+    PyObject * pyarena_create([[maybe_unused]] PyObject *self, PyObject *args)
     {
         int page_size, max_pages;
         if (!PyArg_ParseTuple(
@@ -31,7 +31,7 @@ namespace arenas {
         return pyarena;
     }
 
-    PyObject * pyarena_alloc(PyObject *self, PyObject *args)
+    PyObject * pyarena_alloc([[maybe_unused]] PyObject *self, PyObject *args)
     {
         PyObject *pyArena;
         int element_size;
@@ -79,7 +79,7 @@ arena_t arena_fromcapsule(PyObject *p)
     arena_t arena = (arena_t)PyCapsule_GetPointer(p, OBTYPE_ARENA);
     if (NULL == arena)
     {
-        fprintf(stderr, "%p : AR : BAD\n", p);
+        fprintf(stderr, "%p : AR : BAD\n", reinterpret_cast<void *>(p));
     }
 
     return arena;
--- a/fract4d/c/fract4dc/arenas.cpp
+++ b/fract4d/c/fract4dc/arenas.cpp
@@ -79,7 +79,7 @@ arena_t arena_fromcapsule(PyObject *p)
     arena_t arena = (arena_t)PyCapsule_GetPointer(p, OBTYPE_ARENA);
     if (NULL == arena)
     {
-        fprintf(stderr, "%p : AR : BAD\n", reinterpret_cast<void *>(p));
+        fprintf(stderr, "%p : AR : BAD\n", static_cast<void *>(p));
     }
 
     return arena;
--- a/fract4d/c/fract4dc/functions.cpp
+++ b/fract4d/c/fract4dc/functions.cpp
@@ -183,7 +183,7 @@ ffHandle * ff_fromcapsule(PyObject *pyff)
     ffHandle *ff = (ffHandle *)PyCapsule_GetPointer(pyff, OBTYPE_FFH);
     if (NULL == ff)
     {
-        fprintf(stderr, "%p : FF : CTOR\n", static_cast<void *>(ff));
+        fprintf(stderr, "%p : FF : BAD\n", static_cast<void *>(ff));
     }
     return ff;
 }
--- a/fract4d/c/fract4dc/functions.cpp
+++ b/fract4d/c/fract4dc/functions.cpp
@@ -183,7 +183,7 @@ ffHandle * ff_fromcapsule(PyObject *pyff)
     ffHandle *ff = (ffHandle *)PyCapsule_GetPointer(pyff, OBTYPE_FFH);
     if (NULL == ff)
     {
-        fprintf(stderr, "%p : FF : CTOR\n", reinterpret_cast<void *>(ff));
+        fprintf(stderr, "%p : FF : CTOR\n", static_cast<void *>(ff));
     }
     return ff;
 }
--- a/fract4d/c/fract4dc/functions.cpp
+++ b/fract4d/c/fract4dc/functions.cpp
@@ -16,7 +16,7 @@
 
 namespace functions {
 
-    PyObject * ff_create(PyObject *self, PyObject *args)
+    PyObject * ff_create([[maybe_unused]] PyObject *self, PyObject *args)
     {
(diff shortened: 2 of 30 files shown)


fract4d/c/fract4dc/arenas.cpp
```diff
@@ -79,7 +79,7 @@ arena_t arena_fromcapsule(PyObject *p)
     arena_t arena = (arena_t)PyCapsule_GetPointer(p, OBTYPE_ARENA);
     if (NULL == arena)
     {
-        fprintf(stderr, "%p : AR : BAD\n", p);
+        fprintf(stderr, "%p : AR : BAD\n", reinterpret_cast<void *>(p));
```
DX-MON, line 82: Avoid reinterpret_cast<>() - it's basically almost always UB. In this case, it looks like you can static_cast<>() which will still fix the warning but not err on invoking Undefined Behaviour

fract4d/c/model/STFractWorker.cpp
```diff
@@ -637,6 +633,33 @@ int STFractWorker::diff_colors(rgba_t a, rgba_t b)
...
+        pos.n,
+        options.maxiter,
+        periodGuess(),
+        options.period_tolerance,
+        options.warp_param,
+        x, y, 0,
+        &tpixel, &titer, &tindex, &tfate);
+    if (tpixel == pixel)
+    {
+        m_stats.s[PIXELS_SKIPPED_RIGHT]++;
```
DX-MON, line 654: Prefer `++m_stats.s[PIXELS_SKIPPED_RIGHT];` - compiler will have less work to do and debug will run faster.

fract4d/c/fract4dc/arenas.cpp
mindhells: done! thanks for the feedback

setup.py
```diff
@@ -30,45 +32,37 @@
...
-for path in "/usr/include/jpeglib.h", "/usr/local/include/jpeglib.h":
-    if os.path.isfile(path):
-        extra_macros.append(('JPG_ENABLED', 1))
-        jpg_libs = [jpg_lib]
-        break
-else:
-    raise SystemExit("NO JPEG HEADERS FOUND, you need to install libjpeg-dev")
+jpeg_flags = call_package_config("libjpeg", "--cflags")
+jpeg_libs = call_package_config("libjpeg", "--libs")
+if jpeg_flags + jpeg_libs:
```
cjmayo, line 64: call_package_config() will exit if the package isn't found. Is the concern that jpeg_libs could be empty (that would be a broken system)? In that case suggest changing to a test for `not jpeg_libs` that exits with an error rather than silently proceeding. N.B. PNG_ENABLED above as well.

If both jpeg and png are not optional (as suggested by the call_package_config with optional=False) can the macros be removed?

mindhells: Good catch,
png was already but it's my mistake making jpg mandatory too.

As it is now just a part of gnofract4d, it could make sense to remove macros for mandatory includes. Anyway if the png lib is not found it won't even try to compile fract4dc.
On the other hand, if you look at the C++ layer as something that could be reused elsewhere, it still makes sense to have them (though you could find a better approach).

mindhells:
btw, the idea of enabling the -Wpedantic flag I took it from #134 
just wanted to leave the reference here

setup.py
```diff
@@ -9,9 +9,11 @@
 
 gnofract4d_version = '4.2'
 
+_DEBUG = False
+
 if sys.version_info < (3, 6):
     print("Sorry, you need Python 3.6 or higher to run Gnofract 4D.")
-    print("You have version %s. Please upgrade." % sys.version)
+    print("You have version {sys.version}. Please upgrade.")
```
edyoung, line 16: I actually deliberately left f-strings out of setup.py, because with this change you won't see the error message about python version if you try to run with py3.5, you'll get a parse error instead. Would prefer to leave this file alone but f-strings anywhere else in the code are fine.

fract4d/c/fract4dc/functions.cpp
```diff
@@ -183,7 +183,7 @@ ffHandle * ff_fromcapsule(PyObject *pyff)
     ffHandle *ff = (ffHandle *)PyCapsule_GetPointer(pyff, OBTYPE_FFH);
     if (NULL == ff)
```
edyoung, line 184: This bit doesn't make a lot of sense. We only log the value if it is null? (I have no recollection of this code and don't know if it is being used)

I would follow same pattern as elsewhere with DEBUG_CREATION ifdef

fract4d/c/model/STFractWorker.cpp
```diff
@@ -762,13 +752,13 @@ inline void STFractWorker::rectangle_with_iter(
...
-            m_stats.s[PIXELS]++;
-            m_stats.s[PIXELS_SKIPPED]++;
+            ++m_stats.s[PIXELS];
+            ++m_stats.s[PIXELS_SKIPPED];
         }
     }
 }
 
-// @TODO: this is used for RENDER_THREE_D mode and seems to be unfinished or at leats worth to review
+// @TODO: this is used for RENDER_THREE_D mode and seems to be unfinished or at least worth to review
```
edyoung, line 761: yeah, never finished RENDER_THREE_D

setup.py
edyoung: I don't mind either way whether we make jpg mandatory

edyoung:
Looks good to me, thanks! Don't personally see any advantage in preincrement rather than postincrement but don't object to making that house style

DX-MON:
So, for integers on x86_64 and so.. you won't see a difference because the compilers are really good at eliding away the difference - but for other architectures and more complex types, the difference is this:

With post-increment, you are specifically saying to the compiler "increment thing, but I want to get the value of it first" - so, strictly speaking, the compiler first has to copy the value somewhere, then it can increment the original and store it back to memory. For complex types to increment or types with expensive constructors or assignment overloads, this could be a lot of extra work.

pre-increment avoids this by saying "increment then get me the value", which is of course the natural computational path with no copying the object to retain the object's original value to return.

setup.py
mindhells: I got somewhat carried away :D

fract4d/c/model/STFractWorker.cpp
mindhells: That might be another enhancement to add to [the list](https://github.com/fract4d/gnofract4d/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement).
I'm also very curious about the optimizations I wrapped with EXPERIMENTAL_OPTIMIZATIONS macro in this PR.

PR #149, opened 2020-07-11
Release 4.3

edyoung:
Seems like it is about time for a 4.3 release. 

The only thing I think is currently blocking is needing clearer instructions for how to install in a way that puts all the files in the right places. Please mention anything else you think should go in first.

Changelog:
- Minimum Python version now 3.6
- Supports .ASE gradient files
- Improved manual
- MacOS improvements
- improved icons
- much cleanup of python and C++ code
- better testing infra via GitHub Actions
- removed fractutils folder
- improve install instructions


edyoung:
@mindhells if you think the undefined behavior fixes are nearly ready I would like to wait for those. Otherwise I think we can release.

mindhells:
> @mindhells if you think the undefined behavior fixes are nearly ready I would like to wait for those. Otherwise I think we can release.

I left a comment in #154. I think it's ok to merge but I will continue researching about the GIL, GTK and ref counting playing together.

PR #150, opened 2020-07-11
Edyoung/coverage

edyoung:
Tweak tests to improve coverage


fract4dgui/preferences.py
```diff
@@ -1,11 +1,10 @@
 # GUI for user settings
 
+from . import dialog, utils
+from gi.repository import Gtk, GObject
```
cjmayo, line 4: I would revert this move. For style with the project imports coming last, plus `gi.repository` needs gi and `require_version` comes before Gtk.

edyoung: I don't think I did that on purpose. Maybe autopep8? not sure

PR #151, opened 2020-07-12
Remove unsupported Python 3.5 from docker

cjmayo:
Python 3.6 is the minimum since:
82522d4 ("update min python to 3.6  (#143)", 2020-07-03)


edyoung:
FYI #147  contains very similar changes.

cjmayo:
> FYI #147 contains very similar changes.

Thanks, I'll close this one then.

Issue #73, opened 2020-04-05
Intermittent crash in travis regression tests


edyoung:
Some jobs such as https://travis-ci.com/github/edyoung/gnofract4d/jobs/314557671 are intermittently failing with the stack below. This doesn't appear to be an error in ir.py, which is pure python code and shouldn't cause a segfault. More likely this is a refcount issue, buffer overflow, or similar.

```
fract4dgui/tests/test_browser.py ..Fatal Python error: Segmentation fault

Current thread 0x00007fdd1d9cc680 (most recent call first):

  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/ir.py", line 13 in __init__
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/ir.py", line 57 in __init__
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/ir.py", line 92 in __init__
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 912 in binop
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 779 in exp
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 501 in stm
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 711 in decl
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 491 in stm
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 481 in <listcomp>
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 481 in stmlist
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 155 in final
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 1288 in final
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 1311 in main
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/translate.py", line 1239 in __init__
  File "/home/travis/build/edyoung/gnofract4d/fract4d_compiler/fc.py", line 447 in get_formula
  File "/home/travis/build/edyoung/gnofract4d/fract4d/formsettings.py", line 324 in set_formula
[...]

PR #154, opened 2020-07-13
fract4dc sanitize

mindhells:
Address issues detected with UBSAN ASAN TSAN

I've already managed to find one: https://github.com/fract4d/gnofract4d/issues/73#issuecomment-657593091

Having #133 would help thanks to [Menson ability to add sanitizers out of the box](https://github.com/fract4d/gnofract4d/pull/133#issuecomment-655464733)

### DIFF
--- a/fract4d/c/fract4dc/calcargs.cpp
+++ b/fract4d/c/fract4dc/calcargs.cpp
@@ -1,4 +1,5 @@
 #include "Python.h"
+#include <mutex>
 
 #include "calcargs.h"
 
@@ -17,10 +18,13 @@ calc_args::calc_args()
     params = new double[N_PARAMS];
 }
 
+static std::mutex ref_count_mutex;
+
 void calc_args::set_cmap(PyObject *pycmap_)
 {
     pycmap = pycmap_;
     cmap = colormaps::cmap_fromcapsule(pycmap);
+    const std::lock_guard<std::mutex> lock(ref_count_mutex);
     Py_XINCREF(pycmap);
 }
 
@@ -28,6 +32,7 @@ void calc_args::set_pfo(PyObject *pypfo_)
 {
     pypfo = pypfo_;
     pfo = (loaders::pf_fromcapsule(pypfo))->pfo;
+    const std::lock_guard<std::mutex> lock(ref_count_mutex);
     Py_XINCREF(pypfo);
 }
 
@@ -35,12 +40,14 @@ void calc_args::set_im(PyObject *pyim_)
 {
     pyim = pyim_;
     im = images::image_fromcapsule(pyim);
+    const std::lock_guard<std::mutex> lock(ref_count_mutex);
     Py_XINCREF(pyim);
 }
 void calc_args::set_site(PyObject *pysite_)
 {
     pysite = pysite_;
     site = sites::site_fromcapsule(pysite);
+    const std::lock_guard<std::mutex> lock(ref_count_mutex);
     Py_XINCREF(pysite);
 }
 
@@ -50,6 +57,7 @@ calc_args::~calc_args()
 #ifdef DEBUG_CREATION
     fprintf(stderr, "%p : CA : DTOR\n", this);
 #endif
+    const std::lock_guard<std::mutex> lock(ref_count_mutex);
     Py_XDECREF(pycmap);
     Py_XDECREF(pypfo);
     Py_XDECREF(pyim);
--- a/fract4d/c/fract4dc/calcs.cpp
+++ b/fract4d/c/fract4dc/calcs.cpp
@@ -1,5 +1,6 @@
 #include "Python.h"
-#include <pthread.h>
+#include <thread>
+#include <iostream>
 
 #include "calcs.h"
 
@@ -45,26 +46,11 @@ namespace calcs {
 
         if (cargs->asynchronous)
         {
-            cargs->site->interrupt();
-            cargs->site->wait();
-
-            // cargs->site->start(cargs);
-            cargs->site->start();
-
-            pthread_t tid;
-
-            /* create low-priority attribute block */
-            pthread_attr_t lowprio_attr;
-            //struct sched_param lowprio_param;
-            pthread_attr_init(&lowprio_attr);
-            //lowprio_param.sched_priority = sched_get_priority_min(SCHED_OTHER);
-            //pthread_attr_setschedparam(&lowprio_attr, &lowprio_param);
-
-            /* start the calculation thread */
-            pthread_create(&tid, &lowprio_attr, calculation_thread, (void *)cargs);
-            assert(tid != 0);
-
-            cargs->site->set_tid(tid);
+            auto &site = *cargs->site;
+            site.interrupt();
+            site.wait();
+            site.start();
+            site.set_thread(std::thread(calculation_thread, cargs));
         }
         else
         {
@@ -91,14 +77,11 @@ namespace calcs {
 }
 
 
-void * calculation_thread(void *vdata)
+void * calculation_thread(calc_args *args)
 {
-    calc_args *args = (calc_args *)vdata;
-
 #ifdef DEBUG_THREADS
-    fprintf(stderr, "%p : CA : CALC(%d)\n", args, pthread_self());
+    std::cerr << args << " : CA : CALC(" << std::this_thread::get_id() << ")\n";
 #endif
-
     calc(
         args->options,
         args->params,
@@ -108,11 +91,9 @@ void * calculation_thread(void *vdata)
         args->im,
         0 // debug_flags
     );
-
 #ifdef DEBUG_THREADS
-    fprintf(stderr, "%p : CA : ENDCALC(%d)\n", args, pthread_self());
+    std::cerr << args << " : CA : ENDCALC(" << std::this_thread::get_id() << ")\n";
 #endif
-
     delete args;
     return NULL;
 }
--- a/fract4d/c/model/site.cpp
+++ b/fract4d/c/model/site.cpp
@@ -93,7 +93,6 @@ void FDSite::status_changed(int status_val)
 bool FDSite::is_interrupted()
 {
     //fprintf(stderr,"int: %d\n",interrupted);
-    const std::lock_guard<std::mutex> lock(interrupt_lock);
     return interrupted;
 }
 
@@ -118,12 +117,10 @@ void FDSite::interrupt()
 #ifdef DEBUG_THREADS
     std::cerr << this << " : CA : INT(" << m_thread.get_id() << ")\n";
 #endif
-    const std::lock_guard<std::mutex> lock(interrupt_lock);
     interrupted = true;
 }
 
 void FDSite::start() {
-    const std::lock_guard<std::mutex> lock(interrupt_lock);
     interrupted = false;
 }
 
--- a/fract4d/c/model/site.cpp
+++ b/fract4d/c/model/site.cpp
@@ -33,11 +33,10 @@ void IFractalSite::wait()
 
 inline void FDSite::send(msg_type_t type, int size, void *buf)
 {
-    pthread_mutex_lock(&write_lock);
+    const std::lock_guard<std::mutex> lock(write_lock);
     write(fd, &type, sizeof(type));
     write(fd, &size, sizeof(size));
     write(fd, buf, size);
-    pthread_mutex_unlock(&write_lock);
 }
 
 FDSite::FDSite(int fd_) : fd(fd_), interrupted(false)
@@ -45,7 +44,6 @@ FDSite::FDSite(int fd_) : fd(fd_), interrupted(false)
 #ifdef DEBUG_CREATION
     fprintf(stderr, "%p : FD : CTOR\n", this);
 #endif
-    pthread_mutex_init(&write_lock, NULL);
 }
 
 void FDSite::iters_changed(int numiters)
@@ -61,7 +59,7 @@ void FDSite::tolerance_changed(double tolerance)
 // we've drawn a rectangle of image
 void FDSite::image_changed(int x1, int y1, int x2, int y2)
 {
-    if (!interrupted)
+    if (!is_interrupted())
     {
         int buf[4] = {x1, y1, x2, y2};
         send(IMAGE, sizeof(buf), &buf[0]);
@@ -70,7 +68,7 @@ void FDSite::image_changed(int x1, int y1, int x2, int y2)
 // estimate of how far through current pass we are
 void FDSite::progress_changed(float progress)
 {
-    if (!interrupted)
+    if (!is_interrupted())
     {
         int percentdone = (int)(100.0 * progress);
         send(PROGRESS, sizeof(percentdone), &percentdone);
@@ -79,7 +77,7 @@ void FDSite::progress_changed(float progress)
 
 void FDSite::stats_changed(pixel_stat_t &stats)
 {
-    if (!interrupted)
+    if (!is_interrupted())
     {
         send(STATS, sizeof(stats), &stats);
     }
@@ -95,6 +93,7 @@ void FDSite::status_changed(int status_val)
 bool FDSite::is_interrupted()
 {
     //fprintf(stderr,"int: %d\n",interrupted);
+    const std::lock_guard<std::mutex> lock(interrupt_lock);
     return interrupted;
 }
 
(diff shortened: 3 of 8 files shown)


DX-MON:
That's a pretty nice way to solve the problem.. there's a certain elegance to that vs starting threads paused

mindhells:
> That's a pretty nice way to solve the problem.. there's a certain elegance to that vs starting threads paused

There's a 3rd approach: using pthread_self inside `calculation_thread` and setting `site->set_tid(0)` before creating the thread.

I was wondering about the possibility of `site->wait` being called after `pthread_create` and before `site->set_tid(tid)`. Since the site object is reused by fractal.py that seems to be a possibility but I assume we are protected by the python [global interpreter lock](https://wiki.python.org/moin/GlobalInterpreterLock), are we?

DX-MON:
This feels like something that would benefit from std::thread and I agree that calling set_tid() in the thread itself would be a far better idea as that removes all data races from that code path

mindhells:
> This feels like something that would benefit from std::thread and I agree that calling set_tid() in the thread itself would be a far better idea as that removes all data races from that code path

Tried with std::thread and code looks neater. Not sure I'm doing the canonical way though.
On the other hand, if there's consensus about using this approach, we might need to change `threadpool` as well, to keep consistency.

fract4d/c/fract4dc/calcs.cpp
```diff
@@ -47,24 +48,8 @@ namespace calcs {
...
-            pthread_attr_init(&lowprio_attr);
-            //lowprio_param.sched_priority = sched_get_priority_min(SCHED_OTHER);
-            //pthread_attr_setschedparam(&lowprio_attr, &lowprio_param);
-
-            /* start the calculation thread */
-            pthread_create(&tid, &lowprio_attr, calculation_thread, (void *)cargs);
-            assert(tid != 0);
-
-            cargs->site->set_tid(tid);
+            cargs->site->set_thread(std::thread(calculation_thread, cargs));
```
DX-MON, line 52: This feels like, for reliability, it shouldn't be called via the cargs pointer, IE this:
```
const auto &site = *cargs->site;
site.set_thread(std::thread{calculation_thread, cargs});
```

This might also help the compiler with the other 3 calls to functions on site to do the same for them as it gets rid of a double-indirection..
Otherwise this looks way better!

mindhells: good point! (site couldn't be const though)

DX-MON: ah, fair enough.. my default is to start const and remove as necessary :)

mindhells:
I have this interesting output from the [thread sanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html):

```bash
SUMMARY: ThreadSanitizer: data race object.h:459 in _Py_INCREF(_object*)
==================
==================
WARNING: ThreadSanitizer: data race (pid=67153)
  Write of size 8 at 0x0001153544e0 by main thread:
    #0 _Py_INCREF(_object*) object.h:459 (fract4dc.cpython-38-darwin.so:x86_64+0x2c0b1)
    #1 _Py_XINCREF(_object*) object.h:532 (fract4dc.cpython-38-darwin.so:x86_64+0x2b9eb)
    #2 calc_args::set_cmap(_object*) calcargs.cpp:24 (fract4dc.cpython-38-darwin.so:x86_64+0x2b979)
    #3 parse_calc_args(_object*, _object*) calcs.cpp:172 (fract4dc.cpython-38-darwin.so:x86_64+0x12c31)
    #4 calcs::pycalc(_object*, _object*, _object*) calcs.cpp:41 (fract4dc.cpython-38-darwin.so:x86_64+0x10ef1)
    #5 pycalc(_object*, _object*, _object*) fract4dmodule.cpp:147 (fract4dc.cpython-38-darwin.so:x86_64+0x34d0)
    #6 cfunction_call_varargs <null> (Python:x86_64+0x1a4c1)

  Previous write of size 8 at 0x0001153544e0 by thread T11:
    #0 _Py_DECREF(char const*, int, _object*) object.h:470 (fract4dc.cpython-38-darwin.so:x86_64+0x2c168)
    #1 _Py_XDECREF(_object*) object.h:541 (fract4dc.cpython-38-darwin.so:x86_64+0x2bf77)
    #2 calc_args::~calc_args() calcargs.cpp:53 (fract4dc.cpython-38-darwin.so:x86_64+0x2bea0)
    #3 calc_args::~calc_args() calcargs.cpp:48 (fract4dc.cpython-38-darwin.so:x86_64+0x2c018)
    #4 calculation_thread(calc_args*) calcs.cpp:97 (fract4dc.cpython-38-darwin.so:x86_64+0x13947)
    #5 decltype(std::__1::forward<void* (*)(calc_args*)>(fp)(std::__1::forward<calc_args*>(fp0))) std::__1::__invoke<void* (*)(calc_args*), calc_args*>(void* (*&&)(calc_args*), calc_args*&&) type_traits:3530 (fract4dc.cpython-38-darwin.so:x86_64+0x173af)
    #6 void std::__1::__thread_execute<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void* (*)(calc_args*), calc_args*, 2ul>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void* (*)(calc_args*), calc_args*>&, std::__1::__tuple_indices<2ul>) thread:277 (fract4dc.cpython-38-darwin.so:x86_64+0x16e88)
    #7 void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void* (*)(calc_args*), calc_args*> >(void*) thread:287 (fract4dc.cpython-38-darwin.so:x86_64+0x144b0)

  Thread T11 (tid=3087829, finished) created by main thread at:
    #0 pthread_create <null> (libclang_rt.tsan_osx_dynamic.dylib:x86_64h+0x8aad)
    #1 std::__1::__libcpp_thread_create(_opaque_pthread_t**, void* (*)(void*), void*) __threading_support:336 (fract4dc.cpython-38-darwin.so:x86_64+0x14306)
    #2 std::__1::thread::thread<void* (&)(calc_args*), calc_args*&, void>(void* (&)(calc_args*), calc_args*&) thread:303 (fract4dc.cpython-38-darwin.so:x86_64+0x13d39)
    #3 std::__1::thread::thread<void* (&)(calc_args*), calc_args*&, void>(void* (&)(calc_args*), calc_args*&) thread:295 (fract4dc.cpython-38-darwin.so:x86_64+0x13a1a)
    #4 calcs::pycalc(_object*, _object*, _object*) calcs.cpp:53 (fract4dc.cpython-38-darwin.so:x86_64+0x1154f)
    #5 pycalc(_object*, _object*, _object*) fract4dmodule.cpp:147 (fract4dc.cpython-38-darwin.so:x86_64+0x34d0)
    #6 cfunction_call_varargs <null> (Python:x86_64+0x1a4c1)

SUMMARY: ThreadSanitizer: data race object.h:459 in _Py_INCREF(_object*)
==================
```

It's complaining because we're increasing and decreasing the ref count of the same pycmap object in two different threads: the main one and the one created for the calculation. That happens, I guess, because another call to [pycalc](https://github.com/HyveInnovate/gnofract4d/blob/8c9c19668be8a1435333f766aaba983f480895ed/fract4d/c/fract4dc/calcs.cpp#L39) came in, before the thread created in a previous call has ended.

DX-MON:
more importantly, it's complaining because the decref is in the destructor, meaning the object lifetime finishes and at the same time, the object is incref'd.. this feels like maybe there's a mutex missing around the calc_args object?

mindhells:
> more importantly, it's complaining because the decref is in the destructor, meaning the object lifetime finishes and at the same time, the object is incref'd.. this feels like maybe there's a mutex missing around the calc_args object?

Actually there was a missing mutex in there. I had to use a static mutex because it's not the calc_args instance itself causing the problem but multiple instances of calc_args increasing and decreasing the ref_count of the same instance of cmap, pfo, image and site.

I also got this fixed (hopefully):
```bash
WARNING: ThreadSanitizer: data race (pid=69030)
  Read of size 1 at 0x7b28000172b4 by thread T9:
    #0 FDSite::image_changed(int, int, int, int) site.cpp:62 (fract4dc.cpython-38-darwin.so:x86_64+0x43d66)
    #1 fractFunc::image_changed(int, int, int, int) const fractfunc.h:58 (fract4dc.cpython-38-darwin.so:x86_64+0x3760f)
    #2 fractFunc::reset_progress(float) fractfunc.cpp:129 (fract4dc.cpython-38-darwin.so:x86_64+0x39f7d)
    #3 fractFunc::draw(int, int, float, float) fractfunc.cpp:246 (fract4dc.cpython-38-darwin.so:x86_64+0x3dd17)
    #4 fractFunc::draw_all() fractfunc.cpp:170 (fract4dc.cpython-38-darwin.so:x86_64+0x3b65f)
    #5 calc calcfunc.cpp:38 (fract4dc.cpython-38-darwin.so:x86_64+0x34869)
    #6 calculation_thread(calc_args*) calcs.cpp:85 (fract4dc.cpython-38-darwin.so:x86_64+0x13de5)
    #7 decltype(std::__1::forward<void* (*)(calc_args*)>(fp)(std::__1::forward<calc_args*>(fp0))) std::__1::__invoke<void* (*)(calc_args*), calc_args*>(void* (*&&)(calc_args*), calc_args*&&) type_traits:3530 (fract4dc.cpython-38-darwin.so:x86_64+0x178af)
    #8 void std::__1::__thread_execute<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void* (*)(calc_args*), calc_args*, 2ul>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void* (*)(calc_args*), calc_args*>&, std::__1::__tuple_indices<2ul>) thread:277 (fract4dc.cpython-38-darwin.so:x86_64+0x17388)
    #9 void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void* (*)(calc_args*), calc_args*> >(void*) thread:287 (fract4dc.cpython-38-darwin.so:x86_64+0x149b0)

  Previous write of size 1 at 0x7b28000172b4 by main thread:
    [failed to restore the stack]

  Location is heap block of size 152 at 0x7b28000172a0 allocated by main thread:
    #0 operator new(unsigned long) <null>:5353440 (libclang_rt.tsan_osx_dynamic.dylib:x86_64h+0x5492b)
    #1 sites::pyfdsite_create(_object*, _object*) sites.cpp:18 (fract4dc.cpython-38-darwin.so:x86_64+0xacd0)
    #2 pyfdsite_create(_object*, _object*) fract4dmodule.cpp:128 (fract4dc.cpython-38-darwin.so:x86_64+0x35c0)
    #3 cfunction_call_varargs <null>:5353440 (Python:x86_64+0x1a555)

  Thread T9 (tid=3141054, running) created by main thread at:
    #0 pthread_create <null>:5353488 (libclang_rt.tsan_osx_dynamic.dylib:x86_64h+0x8aad)
    #1 std::__1::__libcpp_thread_create(_opaque_pthread_t**, void* (*)(void*), void*) __threading_support:336 (fract4dc.cpython-38-darwin.so:x86_64+0x14806)
    #2 std::__1::thread::thread<void* (&)(calc_args*), calc_args*&, void>(void* (&)(calc_args*), calc_args*&) thread:303 (fract4dc.cpython-38-darwin.so:x86_64+0x14239)
    #3 std::__1::thread::thread<void* (&)(calc_args*), calc_args*&, void>(void* (&)(calc_args*), calc_args*&) thread:295 (fract4dc.cpython-38-darwin.so:x86_64+0x13f1a)
    #4 calcs::pycalc(_object*, _object*, _object*) calcs.cpp:53 (fract4dc.cpython-38-darwin.so:x86_64+0x11a4f)
    #5 pycalc(_object*, _object*, _object*) fract4dmodule.cpp:147 (fract4dc.cpython-38-darwin.so:x86_64+0x39d0)
    #6 cfunction_call_varargs <null>:5353488 (Python:x86_64+0x1a4c1)

SUMMARY: ThreadSanitizer: data race site.cpp:62 in FDSite::image_changed(int, int, int, int)
```

And finally I have this last one identified by threadsanitizer which I run into when closing the program:
```bash
==================
WARNING: ThreadSanitizer: thread leak (pid=69316)
  Thread T4 (tid=3146887, finished) created by main thread at:
    #0 pthread_create <null> (libclang_rt.tsan_osx_dynamic.dylib:x86_64h+0x8aad)
    #1 std::__1::__libcpp_thread_create(_opaque_pthread_t**, void* (*)(void*), void*) __threading_support:336 (fract4dc.cpython-38-darwin.so:x86_64+0x146f6)
    #2 std::__1::thread::thread<void* (&)(calc_args*), calc_args*&, void>(void* (&)(calc_args*), calc_args*&) thread:303 (fract4dc.cpython-38-darwin.so:x86_64+0x14129)
    #3 std::__1::thread::thread<void* (&)(calc_args*), calc_args*&, void>(void* (&)(calc_args*), calc_args*&) thread:295 (fract4dc.cpython-38-darwin.so:x86_64+0x13e0a)
    #4 calcs::pycalc(_object*, _object*, _object*) calcs.cpp:53 (fract4dc.cpython-38-darwin.so:x86_64+0x1193f)
    #5 pycalc(_object*, _object*, _object*) fract4dmodule.cpp:147 (fract4dc.cpython-38-darwin.so:x86_64+0x38c0)
    #6 cfunction_call_varargs <null> (Python:x86_64+0x1a4c1)

  And 1 more similar thread leaks.

SUMMARY: ThreadSanitizer: thread leak __threading_support:336 in std::__1::__libcpp_thread_create(_opaque_pthread_t**, void* (*)(void*), void*)
==================
```

I haven't figured out this one yet

mindhells:
Btw I had [3 in a row pushes](https://github.com/HyveInnovate/gnofract4d/commits/fract4dc-sanitize) passing the tests, that's encouraging in regard of #73

fract4d/c/model/site.cpp
```diff
@@ -96,6 +93,7 @@ void FDSite::status_changed(int status_val)
 bool FDSite::is_interrupted()
 {
     //fprintf(stderr,"int: %d\n",interrupted);
+    const std::lock_guard<std::mutex> lock(interrupt_lock);
```
edyoung, line 96: don't quite understand why you need a mutex here. We're just checking the current value of a bool. Nothing bad happens if we don't detect that we were interrupted until the next time we check.

fract4d/c/fract4dc/calcargs.cpp
```diff
@@ -17,30 +18,36 @@ calc_args::calc_args()
     params = new double[N_PARAMS];
 }
 
+static std::mutex ref_count_mutex;
+
 void calc_args::set_cmap(PyObject *pycmap_)
 {
     pycmap = pycmap_;
     cmap = colormaps::cmap_fromcapsule(pycmap);
+    const std::lock_guard<std::mutex> lock(ref_count_mutex);
```
edyoung, line 27: I feel like the ideal pattern would be if only the main thread ever dealt with Python refcounts and all the calc threads could ignore it. But haven't thought much about how complicated that would be.

fract4d/c/model/site.cpp
mindhells: I think it’s because of the volatile qualifier that threadsanitizer complains.
Didn’t want to remove the qualifier, I’m not sure I’m missing something about the reason to have it. But maybe it makes sense to just remove it.

fract4d/c/fract4dc/calcargs.cpp
mindhells: I’ve been playing around with GIL acquisition to try to address the last threadsanitizer report. No luck so far.
I guess the GTK signals triggering the calculation process out of the interpreter thread had something to do.

mindhells:
You are very welcome. Appreciate so much your feedback. 
Don’t hesitate to leave any comments open-ended or not.

fract4d/c/model/site.cpp
DX-MON: interrupted should most probably be std::atomic<> and that makes the mutex and data race go away. Volatile is not the right language feature in this case as it doesn't mean thread safe and only causes the compiler to always emit read instructions when the variable is referenced, it doesn't make the compiler ensure cache cohearency or atomicity.

fract4d/c/fract4dc/calcargs.cpp
DX-MON: interrupted should most probably be std::atomic<> and that makes the mutex and data race go away. Volatile is not the right language feature in this case as it doesn't mean thread safe and only causes the compiler to always emit read instructions when the variable is referenced, it doesn't make the compiler ensure cache cohearency or atomicity.

mindhells:
I've just made interrupted std::atomic and removed the mutex. Threadsanitizer warning is gone as @DX-MON predicted.
I'd like to keep playing with GIL and find a more robust solution for the ref counting in calc_args. I'll do it in a different PR.

PR #156, opened 2020-07-14
Fix link in the manual

josebailo:
Little fix of a link that is broken in the manual.

To be more precise, [here](https://fract4d.github.io/gnofract4d/manual/#working-with-files), in the description of the formula file. I assumed that the link should redirect to the formula language reference section.


mindhells:
this PR fixes and anchor link in the manual but anchor links for local help (when you open the help from the program menu help->contents don't work)

edyoung:
thanks!

PR #157, opened 2020-07-14
Tidy Dockerfile and add support for Gio.AppInfo.launch_default_for_uri()

cjmayo:
This results in a Docker image that is about 200MB smaller for me (gtk3.0 seemed to be interpreted as something that needed every mono package going!) and supports the new way of launching a browser.


cjmayo:
`libcairo2-dev` is needed for tox to build pygobject, I'll fix later.

cjmayo:
Put `libcairo2-dev` back, now under run_tox.

PR #158, opened 2020-07-15
Improve frontpage website

josebailo:
I've improved a little bit the website:

- Use CDN to load flickity css and js (enhance load speed and less code to manage).
- Remove some style classes with no use.
- Improve design.
- Improve responsive for mobile devices.


edyoung:
thanks!

PR #165, opened 2020-08-10
Change Codecov GitHub Action dependency to v1

cjmayo:
An upgrade to v1.0.12 at the current time.

----

The v1 tag has to be moved manually in the Codecov repository, when we started using Actions it was pointing at an old version. So, this is an upgrade now and hopefully will move with new major releases in future. We haven't been getting the reports added to our PRs - will this fix it?

(And a documentation commit for macOS)

### DIFF
--- a/.github/workflows/test-gnofract4d.yml
+++ b/.github/workflows/test-gnofract4d.yml
@@ -30,7 +30,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Install MacOS packages
+      - name: Install macOS packages
         if: ${{ runner.os == 'macOS' }}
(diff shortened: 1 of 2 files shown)


DX-MON:
I doubt this will fix the lack of reports on PRs - that should be a function of an integration setting on codecov.io - but it doesn't harm all the same..

cjmayo:
Yes, it does look like the Codecov App (or Team Bot) has to be installed in the repository:
https://docs.codecov.io/docs#advanced-usage

mindhells:
@cjmayo Looks like reports are working again: https://github.com/fract4d/gnofract4d/pull/166#issuecomment-674545476

Issue #164, opened 2020-08-05
Segfaults in version 4.3


jamesjer:
I am the Fedora maintainer of the gnofract4d package.  I am working on updating our package to gnofract4d 4.3, but am seeing segfaults in the test suite.  You can see an example here:

https://koji.fedoraproject.org/koji/taskinfo?taskID=48696332

The segfaults appear to be random; i.e., on this run, it was the arm and ppc64le builders that hit the issue.  On other runs, it is a different architecture or architectures.  The only commonalities I have noted are:

- The segfaults always happen while running fract4dgui/tests/test_browser.py
- The backtrace always points to somewhere in fract4d_compiler, although different places on different runs

jamesjer:
FYI, I tried adding -DDEBUG_ALLOCATION as suggested in https://github.com/fract4d/gnofract4d/issues/109, but it looks like an `#include` is missing:

```
fract4d/c/fract_stdlib.cpp: In function ‘bool arena_add_page(arena_t)’:
fract4d/c/fract_stdlib.cpp:174:13: error: ‘stderr’ was not declared in this scope
  174 |     fprintf(stderr, "%p: ARENA PAGE : CTOR\n", newalloc);
      |             ^~~~~~
fract4d/c/fract_stdlib.cpp:7:1: note: ‘stderr’ is defined in header ‘<cstdio>’; did you forget to ‘#include <cstdio>’?
    6 | #include "fract_stdlib.h"
  +++ |+#include <cstdio>
    7 | 
fract4d/c/fract_stdlib.cpp:174:5: error: ‘fprintf’ was not declared in this scope; did you mean ‘rintf’?
  174 |     fprintf(stderr, "%p: ARENA PAGE : CTOR\n", newalloc);
      |     ^~~~~~~
      |     rintf
fract4d/c/fract_stdlib.cpp: In function ‘s_arena* arena_create(int, int)’:
fract4d/c/fract_stdlib.cpp:195:13: error: ‘stderr’ was not declared in this scope
  195 |     fprintf(stderr, "%p: ARENA : CTOR(%d,%d)\n", arena, page_size, max_pages);
      |             ^~~~~~
fract4d/c/fract_stdlib.cpp:195:13: note: ‘stderr’ is defined in header ‘<cstdio>’; did you forget to ‘#include <cstdio>’?
fract4d/c/fract_stdlib.cpp:195:5: error: ‘fprintf’ was not declared in this scope; did you mean ‘rintf’?
[...]

Issue #80, opened 2020-04-11
Continuous Zoom support


edyoung:
Programs like Xaos provide the ability to hold down a mouse button and zoom continuously into the fractal, as opposed to the click-by-click operation Gnofract 4D currently supports.

mindhells:
We've started to work in this feature. @josebailo has created a [branch](https://github.com/HyveInnovate/gnofract4d/tree/continuous-zoom) in which Gnofract4D allows the user to hold down left/right mouse button to continuously zoom in/out the fractal.
Under the hood, Gnofract4d still does just the same calculations as usual, and the result is not the smooth effect we're looking for. 
In order to get the desired result we need to change the way every "frame" is calculated. [Here](http://web.mit.edu/kolya/sipb/afs/root.afs/athena/activity/p/peckers/OldFiles/Programs/XaoS-2.2/doc/algorithm.txt) Thomas Marsh and Jan Hubicka (original authors of the [XaoS project](https://github.com/xaos-project/XaoS)) explain how they got there. 
In the mentioned article, there's a subtle explanation of the many algorithms they use. We don't know yet if it's possible to bring them to Gnofract4d, due to the special features we have here. Nevertheless, in order to find out we're going to need some reverse engineering to understand their specifics. 

I think the easiest way to get somewhere near the desired result is to follow an iterative process, starting from the most basic approach (maybe some kind of interpolation). To do that, we need to change the way Gnofract4d calculates the pixels and also add new data structures to reuse calculations from the previous frame (which is the foundation of the Xaos algorithms).

At this moment I'm working in the idea of creating a new `IFractWorker` implementation (not sure yet if I will need a new `fractFunc` as well) to be the place to implement the new algorithms. Hopefully I'll manage to keep the interface to ease switching between normal and XaoS zooming.

mindhells:
1st approach created by @josebailo is based on a Timer that zooms and calculates the frame every 300ms while the user holds the button. Problem with this is you don't know exactly how much time you need to calculate the previous frame, but you need the previous frame to be calculated. We need then to sync the animation frequency with the previous frame calculation ending.

On the other hand, while this is just a proof of concept for now, we need to define some requirements of what the final result should look like. As some of the current features could clash with it (the selection rect for example), I see this new feature as a different "working mode", like the user has to toggle some sort of switch to enable it (and disable other features)... or maybe the user needs to hold a key. What would be the better, or more consistent, approach?

mindhells:
[...]

PR #166, opened 2020-08-16
Hold the GIL while deleting python objects

jamesjer:
This code is in a thread not created by python, so it does not acquire the GIL by default.  The delete statement in question drops references to multiple python objects, possibly causing them to be deallocated while not holding the GIL.  This appears to be the direct cause of https://github.com/fract4d/gnofract4d/issues/164.

### DIFF
--- a/fract4d/c/fract4dc/calcs.cpp
+++ b/fract4d/c/fract4dc/calcs.cpp
@@ -94,7 +94,9 @@ void * calculation_thread(calc_args *args)
 #ifdef DEBUG_THREADS
     std::cerr << args << " : CA : ENDCALC(" << std::this_thread::get_id() << ")\n";
 #endif
+    PyGILState_STATE gstate = PyGILState_Ensure();
     delete args;
+    PyGILState_Release(gstate);
     return NULL;
 }
 


mindhells:
This is very interesting, and very likely related to the errors we were trying to address in #73. I myself [tried to find out if GIL](https://github.com/fract4d/gnofract4d/pull/154#issuecomment-660675974) had something to do, but I don't have much experience with the Python C-API.

About your approach, it looks good to me. I might be interesting to wrap it in within a class to take advantage of RAII though. [Something like](https://stackoverflow.com/questions/651048/concurrency-are-python-extensions-written-in-c-c-affected-by-the-global-inter):
```C++
      struct ensure {
            PyGILState_STATE* state;
            bool active;

            ensure()
                :state(PyGILState_Ensure()), active(true)
                {}

            ~ensure() { if (active) { restore(); } }

            void restore() {
                PyGILState_Release(state);
                active = false;
            }
        };
```
The [official docs](https://docs.python.org/3/c-api/init.html#c.PyGILState_Ensure) recommend to check if the interpreter is finalizing. Don't know if that's worth in this case.

mindhells:
As you can see in #80, we are working in a continuous zoom feature. This feature involves calling this fract4dc extension API with high frequency. We've been having some trouble with segfault errors when increasing the framerate so I decided to try adding this fix. I must tell stability has improved a lot: no segfault errors ever since.

So, from my side I would approve and merge this changes. We could add the RAII improvements afterwards, if they add some value.

mindhells (approved):

DX-MON (approved):

DX-MON:
Seems good to me though I would prefer to see it use the RAII approach.. but for right now, this seems very acceptable.

PR #170, opened 2020-09-02
Stop skipping testFDSite

cjmayo:
Recent fixes to fract4dc have resolved intermittent test failures.

---

Let's see how the check goes... No, I am confident I have run this multiple times.

(Fantastic to have those fixes in - really felt like the last bit of the jigsaw after all the work to bring Gnofract4D up-to-date).

### DIFF
--- a/fract4d/tests/test_fract4d.py
+++ b/fract4d/tests/test_fract4d.py
@@ -4,7 +4,6 @@ import os.path
 import struct
 import math
 import select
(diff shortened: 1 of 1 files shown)


cjmayo:
At first look appears that the call to random() in fract4d/gradient.py is causing the variation in coverage.

DX-MON (approved):

PR #171, opened 2020-09-03
Test on Python 3.8 and 3.9

cjmayo:
I've tried to minimise the number of checks.

Unfortunately it seems you do have to give the full version for pre-release versions of Python.


cjmayo:
3.9 failed on testFDSite, enabled in a6b74ac16ab727bb87113e40ce84e212968ae713. I hope that is just for 3.9, and merging that hasn't caused more problems (I haven't seen it fail anywhere else).

edyoung:
Thanks!

cjmayo:
The testFDSite failures aren't just Python 3.9, so let's go for it!

PR #172, opened 2020-09-03
Fix inconsistent test coverage due to use of random()

cjmayo:
Directly test RGBtoHSV() and HSVtoRGB().
Use a formula with all parameter types and mock random() to fully test
formsettings mutate().

---

Based on the codecov results from #170, maybe there are others.


DX-MON (approved):

edyoung (approved):
thanks!

PR #173, opened 2020-09-04
Make testFDSite more robust by limiting message bytes read

cjmayo:
Copied from fract4dgui.gtkfractal.Hidden.onData().

---

This does fix some problems, not convinced it does everything though.

onData seems to just discard messages of the wrong size:

https://github.com/fract4d/gnofract4d/blob/5aaad2ae1d98bada86b0f16d3b3cbaff51150c4c/fract4dgui/gtkfractal.py#L175-L186

I don't know if anyone has any better insight. Else, if the test failures are blocking other work then going back to skipping by reverting a6b74ac16ab7 would probably be best.


cjmayo:
FWIW I have never seen the failure on Gentoo (Python 3.8) only Ubuntu (in a VM, Gentoo in a VM seems fine too - I have tested that a bit more then Ubuntu in a VM but not as much as native Gentoo).

PR #174, opened 2020-09-09
Fix more inconsistent test coverage due to use of random()

cjmayo:
Some more tests to stop fluctuating coverage. And fix what looks like a bug where angle_random() was never returning -pi/2:

https://github.com/fract4d/gnofract4d/blob/5aaad2ae1d98bada86b0f16d3b3cbaff51150c4c/fract4d/fractal.py#L653-L658


DX-MON (approved):
Looks good - nicely found with that missing minus on the angle_random() division constant

PR #175, opened 2020-09-17
[RFC] Continuous zoom

mindhells:
Please see #80 to better understand what we are trying to do here.

### DIFF
--- a/doc/continuous_zooming.md
+++ b/doc/continuous_zooming.md
@@ -4,9 +4,9 @@ Continuous zooming feature allows the user to dynamically explore the fractal by
 
 ## User experience
 
-While the CZ feature is running, a fixed amount of zoom is performed in/out depending on which button the user is holding (left/right). The framerate is fixed, as well as the amount of zooming, but since some computers could take longer than the time expected to render a single frame, the frame rate might drop and not be constant. As it might be already obvious, the higher the resolution the more chances for the frame rate to drop, always depending on the CPU power.
+While the CZ feature is running, a fixed amount of zoom is performed in/out depending on which button the user is holding (left/right). The frame rate is fixed, as well as the amount of zooming, but since some computers could take longer than the time expected to render a single frame, the frame rate might drop and not be constant. As it might be already obvious, the higher the resolution the more chances for the frame rate to drop, always depending on the CPU power.
 
-CZ uses different algorithms to enhance performance and get a decent framerate, including _pixel reutilization_ or _dynamic resolution_ which are explained further below in this document. Nevertheless, since Gnofract 4D is highly customizable and allows the user the create his own fractals, the computing power needed to render each frame could vary a lot.
+CZ uses different algorithms to enhance performance and get a decent frame rate, including _pixel reutilization_ or _dynamic resolution_ which are explained further below in this document. Nevertheless, since Gnofract 4D is highly customizable and allows the user the create his own fractals, the computing power needed to render each frame could vary a lot.
 
 Some of the regular preferences or features of Gnofract 4D do not take effect when working in CZ mode. That includes _auto deepen_, _auto tolerance_ and _antialiasing_.
 
@@ -38,7 +38,7 @@ With this basic idea in mind, and in a few words, Gnofract 4D sends all the para
 
 While the communication scheme subtlely explained above is fine for the regular functioning mode, it presents some drawbacks for the CZ mode.
 
-In the regular mode, every time a new fractal image has to be drawn, C++ performs some boilerplate initializations. Doing this for an on-going animation that creates an new image every few miliseconds is unbearable.
+In the regular mode, every time a new fractal image has to be drawn, C++ performs some boilerplate initializations. Doing this for an on-going animation that creates a new image every few miliseconds is unbearable.
 
 For that reason, scheme communication for CZ is slightly different. Instead of using the _regular_ image calculation trigger API from the `fract4dc` extension, a new API has been created with the following conceptual facade:
 - start CZ process: this triggers the image calculation process and perform boilerplate initializations on the C++ layer. It also creates an infinite loop waiting for updates from the python layer.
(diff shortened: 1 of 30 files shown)


edyoung:
Hi folks! Just wanted to say - thanks a ton for all the work that has gone into this! Sorry I have been scarce the last few months. I've been super busy at work. I have some time off over the holidays so will aim to review, merge and release this as soon as we can.

mindhells:
> Hi folks! Just wanted to say - thanks a ton for all the work that has gone into this! Sorry I have been scarce the last few months. I've been super busy at work. I have some time off over the holidays so will aim to review, merge and release this as soon as we can.

That's great news. I'm going to post the last updates in the related issue so you have all the details.
There's a couple of more things to add to this PR before it's ready for review, the idea is to have them before christmas.

doc/continuous_zooming.md
```diff
@@ -0,0 +1,89 @@
...
+
+`Location` (`X`, `Y` ... and `Z`, `W` in 4D space) and `Size` parameters will determine which point of the complex plane correspond to every pixel of the image (we leave out from this explanation the plane rotation parameters). This makes very likely to happen that subsequent frames, where location and size would change slightly, have pixels with the same, or very close point in the complex plane. So it makes sense to, instead of calculate the color of that pixel again (which is the heaviest process) we take the already calculated value.
+
+The process to find those pixels with same or close complex plane point values, in an efficient way, is based in the assumption that _imaginary values `y` remain constant while moving across X axis, while real values `x` do the same for the Y axis_. And it consist in the following steps:
+- find all the `x` values or columns for the new fractal (image) that are close enough to the previous image ones.
```
edyoung, line 59: One way of thinking about this: imagine you are taking the previous frame and scaling it up. In frame 1, to get the color for pixel (0,0) we calculated point (x,y,z,w). In Frame 2, point (x,y,z,w) is now at a different pixel location, but if that location is still on screen, we can reuse the calculation and color pixel(a,b) with that point. I'm pretty sure that doesn't really depend on only x changing as we move across the screen or y changing vertically. We have a vector (dx,dy,dz,dw) which is the amount the point changes when we step 1 pixel to the left - we can use that to generalize this to work for rotated images or ones in other planes.

mindhells: That was exactly my thought when I first read https://github.com/xaos-project/XaoS/wiki/Developer's-Guide#saving-previous-pixels.
Then I said to myself "ok, let's try only XY plane (no plane rotation)". I was surprised when I found out you can actually change plane rotations and it keeps working. Except in certain cases.
As the Gnofract 4D manual states:
> It allows you to treat a fractal which has more than one parameter as a four-dimensional object and interactively view slices of this object from arbitrary angles, giving rise to some very unusual images.
It seems that, as long as the slice you are viewing is perpendicular or parallel to the XY plane the assumption is valid. In other words, if you rotate the slice respect to X and Y at the same time, the it  doesn't work. 
I tried to solve this issue/limitation, being playing around with the rotation matrix, point vectors... but honestly I don't have the math background to.

mindhells:
@edyoung is it ok if we merge this PR? 
We haven't received any additional feedback so far and it would be a shame to leave this here forever. Besides, this feature could be enabled/disabled by configuration.

edyoung:
yes! go for it

da2ce7:
Needs Rebase.

fract4dgui/gtkfractal.py
```diff
@@ -21,7 +21,6 @@
 
 from . import utils, fourway
 
-
```
cjmayo, line 24: Two blank lines surrounding class definitions is from PEP 8 (and fract4dgui does now pass a flake8 check!).

fract4dgui/preferences.py
```diff
@@ -401,6 +401,23 @@ def set_prefs(*args):
 
         return widget
 
+    def create_continuous_zoom_widget(self):
+        widget = Gtk.CheckButton(label="Continuous _Zoom")
+        widget.set_tooltip_text("Activate continuous zoom to zoom in or out while keep clicked the mouse buttons")
+        widget.set_use_underline(True)
```
cjmayo, line 407: tooltip_text and use_underline can be used as keyword arguments in Gtk.CheckButton() as per the neighbouring methods now.

fract4dgui/gtkfractal.py
```diff
@@ -975,9 +1062,14 @@ def filterPaintModeRelease(self, event):
         return False
 
     def onButtonRelease(self, widget, event):
-        self.widget.queue_draw()
+        # self.widget.queue_draw()
```
cjmayo, line 1065: Doesn't appear to be needed, let's get rid of it.

josebailo:
Rebase done!

@cjmayo I removed the commented line and refactored the `create_continuous_zoom_widget`.

Regarding the flake8 check, I haven't found anything about using flake8 in the project so I'm not sure how to proceed. Are you using it by yourself or is established for developing?

mindhells:
Hi @edyoung ,
after rebase, test are [failing](https://github.com/fract4d/gnofract4d/pull/175/checks?check_run_id=2321700604) because this condition is true:
https://github.com/fract4d/gnofract4d/blob/aab8a8f6a863f83f9de508f82e412bd2a51aa580/fract4d/c/model/site.cpp#L24-L29
can you elaborate under what circumstances this may happen? Maybe we're missing something anywhere else

cjmayo:
> Regarding the flake8 check, I haven't found anything about using flake8 in the project so I'm not sure how to proceed. Are you using it by yourself or is established for developing?

It's only me.  I just mentioned it because the single blank line before the class was obvious, that's fixed after the rebase.

josebailo:
Sorry for the delay.

I've been checking why the tests fails. First I thought this was because of some change on this brand, but I found a commit by @edyoung in another branch skipping this test because it fails intermittently.

I've skipped the test and it works now!

PR #182, opened 2020-10-22
Only draw once on application startup

cjmayo:
MainWindow.on_fractal_change() was being called several times while the
first fractal was being created.

Prevents application hangs on startup since:
47752e0 ("Hold the GIL while deleting python objects", 2020-08-16)

---

Please check and test, I haven't looked at this code before and got here by experimentation.

### DIFF
--- a/fract4dgui/main_window.py
+++ b/fract4dgui/main_window.py
@@ -81,7 +81,7 @@ class MainWindow:
         self.window.add(self.vbox)
 
         self.f = gtkfractal.T(self.compiler, self)
-        self.f.freeze()  # create frozen - main prog will thaw us
+        self.f.freeze()  # create frozen - main prog queues first_draw() to thaw
         self.create_subfracts(self.f)
 
         self.set_filename(None)
@@ -441,6 +441,10 @@ class MainWindow:
             title += "*"
 
         self.window.set_title(title)
+
+    def first_draw(self):
+        self.on_fractal_change()
+        self.f.thaw()
 
     def on_fractal_change(self, *args):
         self.draw()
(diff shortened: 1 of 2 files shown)


fract4dgui/main_window.py
```diff
@@ -442,6 +442,10 @@ def set_window_title(self):
 
         self.window.set_title(title)
 
+    def first_draw(self, *args):
+        self.on_fractal_change()
+        self.f.thaw()
```
cjmayo, line 447: Appears that this thaw() is causing a draw and could go (which means first_draw() can go too). e.g. update_subfracts() calls freeze() and thaw().

gnofract4d
```diff
@@ -62,11 +62,7 @@ def gtkmain(userConfig, options):
 
     mainWindow = main_window.MainWindow(userConfig)
     mainWindow.apply_options(options)
-
-    if mainWindow.f.thaw():
-        gi.require_version('GLib', '2.0')
-        from gi.repository import GLib
-        GLib.idle_add(mainWindow.on_fractal_change)
```
cjmayo, line 69: My first attempt used idle_add to call first_draw. Removing it didn't seem to make a difference, but maybe we still want it to keep supporting less powerful CPUs, or just in case there are other demands on the system.

fract4dgui/main_window.py
cjmayo: But, e.g. moving the fractal doesn't call freeze(), thaw() i.e. a thaw() is essential.

gnofract4d
cjmayo: Removing idle_add was the cause of the additional draw on thaw() above.

DX-MON (approved):

PR #184, opened 2020-11-20
Fix compile errors on Mac

mindhells:
Not sure since when this is happening or what has changed, but the last test jobs for mac are failing with this errors:

```bash
error: implicit declaration of function 'gradient' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
```
```bash
error: implicit declaration of function 'image_lookup' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
```
see https://github.com/fract4d/gnofract4d/runs/1325433239 for more details.

Those 2 functions `gradient` and  `image_lookup` are defined in `model\colorutils.h` and `model\image.h` which are not included in the formula .c file created by `codegen.py`. But as far as I know this is not new, they never were.

In previous versions, definition of those functions was in fract4d_stdlib extension, and from this [commit](https://github.com/fract4d/gnofract4d/commit/8514d71853c107fe18e7b64adddf78794e13331a) on is on the fract4d main extension. But still nothing changes. 

I've added `-Wno-implicit-function-declaration` flag to the compiler command which solves the problem.
Another possible solution would be to change `codegen.py` to include the missing headers and also include those header files in the package_data, like in #91


DX-MON:
This works, but I'd honestly go with include the headers for if/when this ends up converted to C++ where implicit definitions are hard errors.

DX-MON (approved):

PR #185, opened 2020-12-11
Correct title of Save High-Res Image shortcut

cjmayo:
I copied the method comment instead of the description for the user from Save High-Res Image.

Also:

Remove unused MainWindow.get_all_actions()

Not used since:
ae9f710 ("Change documentation to markdown (#101)", 2020-06-07)

(fract4dgui/createdocs.py was deleted)


edyoung:
Thanks!

PR #186, opened 2020-12-11
Raise minimum GTK to 3.22 and fix Gdk.Screen deprecation warnings

cjmayo:
GTK 3.22 is present in Ubuntu 18.04 LTS (which we are using in GitHub Actions). Debian stable has GTK 3.24.

What is the minimum OS we want to support? Raising the minimum Python to 3.6 has already ruled out Ubuntu 16.04 LTS (which ends support in April 2021).

GtkShortcutsShortcut:action-name would be useful for some work I am just about to post.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Requirements
 Gnofract 4D requires these packages to run:
 
 - Python version 3.6 or higher
-- GTK version 3.18 or higher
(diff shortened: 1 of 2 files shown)


edyoung:
LGTM. Thanks!

PR #193, opened 2020-12-17
Bugfix for adjusting parameter values

edyoung:
RFC doc ought to be in another PR but oh well

### DIFF
--- a/fract4dgui/utils.py
+++ b/fract4dgui/utils.py
@@ -73,14 +73,6 @@ def set_menu_from_list(menu, items):
     menu.set_model(model)
 
 
-def set_selected(menu, i):
-    menu.set_active(int(i))
-
-
-def get_selected(menu):
-    return menu.get_active()
-
-
 def get_selected_value(menu):
     iter = menu.get_active_iter()
     if not iter:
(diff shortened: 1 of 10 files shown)


fract4dgui/utils.py
```diff
@@ -21,26 +21,10 @@
 break_new_things = False
```
cjmayo, line 21: I think `threads_enabled`, and use in gtkfractal.py, could go then.

`break_new_things` doesn't appear anywhere else either.

fract4dgui/utils.py
```diff
@@ -21,26 +21,10 @@
...
-        Gdk.threads_leave()
-
-
-def idle_wrapper(callable, *args):
-    threads_enter()
-    callable(*args)
-    threads_leave()
-
-
 def idle_add(callable, *args):
```
cjmayo, line 24: `idle_add()` is now just mirroring `GLib.idle_add()` which is already used in PNGGen.py and main_window.py. Suggest it can go as well too - a number in gtkfractal.py, doesn't look like it will clash with #175 at a quick read through.

Either way the comment is now incorrect.

PR #195, opened 2020-12-17
Use actions for arrow keys

cjmayo:
Allows complex event processing to be removed.

---

That's the last GAction change I have.


edyoung (approved):
lgtm

PR #203, opened 2021-01-08
Remove utils.input_add()

cjmayo:
utils.input_add() only used once.

### DIFF
--- a/fract4dgui/gtkfractal.py
+++ b/fract4dgui/gtkfractal.py
@@ -121,7 +121,11 @@ class Hidden(GObject.GObject):
             self.f.saved = val
 
     def input_add(self, fd, cb):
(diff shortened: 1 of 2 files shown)


DX-MON (approved):
I might still have left it as a helper against gtkfractal.py, however this change makes a lot of sense to make. My logic for keeping it `self.input_add(self.readfd, self.onData)` is that call more clearly expresses the intent.. either way though

cjmayo:
> I might still have left it as a helper against gtkfractal.py, however this change makes a lot of sense to make. My logic for keeping it `self.input_add(self.readfd, self.onData)` is that call more clearly expresses the intent.. either way though

OK - I was a bit keen on removing things. I've recreated the helper.

PR #205, opened 2021-01-10
MainWindow parameters

cjmayo:
Simplify MainWindow.__init__().

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -43,8 +43,9 @@ class Benchmark:
 
     def run_gui(self):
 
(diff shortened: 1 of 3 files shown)


cjmayo:
Missed something when I split this from the larger PR. Will fix...

cjmayo:
OK now, tests and benchmark.py fixed.

cjmayo:
Removed `Remove unused extra_paths parameter from MainWindow.__init__()`, that is better done in the update of MainWindow.

DX-MON (approved):

PR #207, opened 2021-01-10
[WIP] Split MainWindow

cjmayo:
I've always found the size of MainWindow, 1300 lines, made it difficult to work on. Here is a suggestion of how to make that easier.

The creation tasks are done in application_window.py. The Actions and ApplicationDialogs classes aren't very smart, just a way to categorise the methods.

application_widgets.py contains a couple of more complete classes, notably FractalWindow.

WIP because it is quite a lot of change and perhaps there are some opinions. I don't think this is perfection or the last word but hopefully it would be a step forward. It also uses the other smaller PRs that I have submitted.


cjmayo:
Ready to go now. I moved the test change commit into the new GUI tests PR (it works on current master).

I realise this is quite a big change, if anyone wants to say wait because they don't have enough time to review right now - please do so.

PR #209, opened 2021-01-18
Fix --antialias segfaults

cjmayo:
Broken in:
cdcd87e ("convert options.py to use argparse", 2018-03-04)


cjmayo:
I'll fix the LGTM.com warning first.

cjmayo:
Warning fixed, by moving the code into fractal.py - a better solution I think, thanks LGTM.com.

PR #217, opened 2021-01-26
Fix duplicating formula_paths in user configuration

cjmayo:
If the user removes _shared_formula_dir, on the next start-up
a second ${HOME}/gnofract4d/formulas was also being added.

Fixed by removing the numeric keys from the path items - order is
preserved by the dictionary containing them.

---

Split out because it changes the config file format - but includes code to automatically convert older files to make it transparent.

This one is draft because it is based on #213.

### DIFF
--- a/fract4d/fractconfig.py
+++ b/fract4d/fractconfig.py
@@ -68,19 +68,19 @@ class T(configparser.ConfigParser):
              OrderedDict()
              ),
             ("formula_path",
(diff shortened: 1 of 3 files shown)


cjmayo:
OK, ready to go now (rebased on master with pre-req merged).

DX-MON (approved):

PR #223, opened 2021-02-02
Tidy GUI folder

cjmayo:
This works through the fract4dgui folder:
* using keyword arguments with Gtk constructors where possible
* replacing deprecated Gtk.Hbox and Gtk.Vbox
* converting instance variables to local variables where possible
* removing `pylint: disable=no-member` where possible
* enabling the whole folder to pass all flake8 tests (except E402 for files with `gi.require_`)

### DIFF
--- a/fract4dgui/angle.py
+++ b/fract4dgui/angle.py
@@ -33,18 +33,16 @@ class T(Gtk.DrawingArea):
 
         self.adjustment.connect('value-changed', self.onAdjustmentValueChanged)
 
-        Gtk.DrawingArea.__init__(self, tooltip_text=tip)
-
-        self.set_size_request(40, 40)
+        Gtk.DrawingArea.__init__(self, tooltip_text=tip, width_request=40, height_request=40)
 
         self.set_events(
-            Gdk.EventMask.BUTTON_RELEASE_MASK |
-            Gdk.EventMask.BUTTON1_MOTION_MASK |
-            Gdk.EventMask.POINTER_MOTION_HINT_MASK |
-            Gdk.EventMask.ENTER_NOTIFY_MASK |
-            Gdk.EventMask.LEAVE_NOTIFY_MASK |
-            Gdk.EventMask.BUTTON_PRESS_MASK |
-            Gdk.EventMask.EXPOSURE_MASK
+            Gdk.EventMask.BUTTON_RELEASE_MASK
+            | Gdk.EventMask.BUTTON1_MOTION_MASK
+            | Gdk.EventMask.POINTER_MOTION_HINT_MASK
+            | Gdk.EventMask.ENTER_NOTIFY_MASK
+            | Gdk.EventMask.LEAVE_NOTIFY_MASK
+            | Gdk.EventMask.BUTTON_PRESS_MASK
+            | Gdk.EventMask.EXPOSURE_MASK
         )
 
         self.notice_mouse = False
@@ -69,7 +67,6 @@ class T(Gtk.DrawingArea):
 
     def angle_from_xy(self, x, y):
         angle = math.atan2(y, x)
-        #angle = math.fmod(angle,T.two_pi)
         return angle
 
     def onAdjustmentValueChanged(self, adjustment):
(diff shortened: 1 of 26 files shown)


fract4dgui/angle.py
```diff
@@ -33,18 +33,16 @@ def __init__(self, text, tip):
...
 
         self.set_events(
-            Gdk.EventMask.BUTTON_RELEASE_MASK |
-            Gdk.EventMask.BUTTON1_MOTION_MASK |
-            Gdk.EventMask.POINTER_MOTION_HINT_MASK |
-            Gdk.EventMask.ENTER_NOTIFY_MASK |
-            Gdk.EventMask.LEAVE_NOTIFY_MASK |
-            Gdk.EventMask.BUTTON_PRESS_MASK |
-            Gdk.EventMask.EXPOSURE_MASK
+            Gdk.EventMask.BUTTON_RELEASE_MASK
```
DX-MON, line 39: I'm intrigued as to your choice here.. the code base consistently puts the |'s on the right at the moment, and I for one find this much easier to read - while the at the beginning not so much.

The consistency point is of more interest to me here though.

fract4dgui/fourway.py
```diff
@@ -19,18 +19,16 @@ def __init__(self, text, tip):
...
 
         self.set_events(
-            Gdk.EventMask.BUTTON_RELEASE_MASK |
-            Gdk.EventMask.BUTTON1_MOTION_MASK |
-            Gdk.EventMask.POINTER_MOTION_HINT_MASK |
-            Gdk.EventMask.ENTER_NOTIFY_MASK |
-            Gdk.EventMask.LEAVE_NOTIFY_MASK |
-            Gdk.EventMask.BUTTON_PRESS_MASK |
-            Gdk.EventMask.EXPOSURE_MASK
+            Gdk.EventMask.BUTTON_RELEASE_MASK
```
DX-MON, line 25: Same comment as before about the consistency of this with the rest of the code base.

fract4dgui/gtkfractal.py
```diff
@@ -463,14 +463,14 @@ def __init__(self, comp, parent=None, width=640, height=480):
...
         drawing_area.set_events(
-            Gdk.EventMask.BUTTON_RELEASE_MASK |
-            Gdk.EventMask.BUTTON1_MOTION_MASK |
-            Gdk.EventMask.POINTER_MOTION_MASK |
-            Gdk.EventMask.POINTER_MOTION_HINT_MASK |
-            Gdk.EventMask.BUTTON_PRESS_MASK |
-            Gdk.EventMask.KEY_PRESS_MASK |
-            Gdk.EventMask.KEY_RELEASE_MASK |
-            Gdk.EventMask.EXPOSURE_MASK
+            Gdk.EventMask.BUTTON_RELEASE_MASK
```
DX-MON, line 466: Ditto the previous two comments.

fract4dgui/settings.py
```diff
@@ -150,18 +147,17 @@ def create_colors_table(self):
...
+        self.gradarea = Gtk.DrawingArea(width_request=256, height_request=96)
 
         self.gradarea.add_events(
-            Gdk.EventMask.BUTTON_RELEASE_MASK |
-            Gdk.EventMask.BUTTON1_MOTION_MASK |
-            Gdk.EventMask.POINTER_MOTION_HINT_MASK |
-            Gdk.EventMask.BUTTON_PRESS_MASK |
-            Gdk.EventMask.KEY_PRESS_MASK |
-            Gdk.EventMask.KEY_RELEASE_MASK
+            Gdk.EventMask.BUTTON_RELEASE_MASK
```
DX-MON, line 153: Same comment as before.

DX-MON (changes requested):

fract4dgui/angle.py
cjmayo: PEP 8 changed (for new code) five years ago on this:
https://github.com/python/peps/commit/7690a00d478866eb3cae3d82bbe1e80199a7392f#diff-db417657713fab81e79cb425b07ca55ecb11fc01fc52627e2756d450e8ac6f04

apparently to follow mathematical convention as recorded by Donald Knuth.

I only changed it because `flake8` was complaining - but I must of had a global setting in place because it is OK now. I'll change them back, I don't want to make it harder to read.

cjmayo: Those are all done now in a new commit.

PR #234, opened 2021-02-18
Allow "Save changes" dialog to be closed by Escape

cjmayo:
One fix for pressing Escape causes the "Save changes" exit dialog to be reshown.

Plus some final tidying up that is the end of this sequence of changes from me.


edyoung:
Chris, I am way behind on reviewing your series of changes. Thanks so much for making them and sorry for the hold up. Please feel free to merge them when you think they are ready, without waiting for sign off. these all sounds like worthy updates. Thanks again!

DX-MON (approved):

cjmayo:
Thanks @edyoung, good to know you are happy with the changes and thanks to @DX-MON for the reviews. Definitely preferable to have the reviews but got to be realistic.

PR #243, opened 2021-11-01
Test with Python 3.10 on GitHub and Upgrade to Codecov Action v2

cjmayo:


.github/workflows/test-gnofract4d.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04]
-        python: [3.8, 3.9]
+        python: [3.8, 3.9, "3.10"]
```
DX-MON, line 16: Does this need the quotes to work or can it be done without? If it needs them, we suggest converting the other values here to use them too, otherwise we'd drop the quotes around 3.10

cjmayo: Without the quotes it tries to use Python 3.1. Hadn't realised that before pushing here. Updated to use quotes around all Python versions.

cjmayo:
Realised that the number of jobs before sending to Codecov needed increasing in codecov.yml - added a comment in test-gnofract4d.yml.

Checked the Codecov docs as a result and found the v1 action is scheduled to stop working in February - with intermittent outages before then. Upgraded to the v2 action.

DX-MON (approved):

DX-MON:
Once the tests complete, I'll merge. Happy with this.

PR #244, opened 2021-11-01
Fix bdist_rpm after recent changes

cjmayo:
Fails because COPYING cannot be found and does not include new
resources.


DX-MON (approved):
Looks good to me

cjmayo:
Should have mentioned that doc-files in setup.cfg was duplicating installation already done explicitly in setup(),

PR #247, opened 2022-10-28
Testing Updates

cjmayo:
Test with Python 3.11

GitHub Actions have deprecated Node 12 and macOS 10

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/


dragonmux (approved):
LGTM, merging. Had to go through similar pain on another repo

PR #248, opened 2022-10-28
Fix tests failing with pytest due to missing _()

cjmayo:
TypeError: 'NoneType' object is not callable

---

Something changed when using pytest on its own - I've seen this on other projects. You now have to have the gettext `_()` alias installed.


dragonmux (approved):
This seems reasonable considering.. Merging.

PR #249, opened 2022-11-08
Use Gio.ActionMap.add_action_entries()

cjmayo:
Batches up the actions instead of calling add_action() multiple times.

Suggested in bug for missing override https://gitlab.gnome.org/GNOME/pygobject/-/issues/426.


edyoung (approved):
LGTM. Thanks!

Issue #14, opened 2018-02-04
dragging selection rectangle slow and laggy in python3


edyoung:
click and drag to draw rectangle (python3 branch). it can't keep up with the mouse

cjmayo:
Issue introduced with conversion to GTK 3 and cairo resolved in conversion to GTK 4 and render nodes.

PR #253, opened 2023-04-10
Convert to GTK 4

cjmayo:
One clear advantage of GTK 4 for Gnofract 4D is that it makes it easy to fix the high CPU usage while dragging the selection rectangle #14. With the new scene graph a widget's render node is cached. We can split the selection and fractal into separate widgets and only update the fractal widget when there are actual changes. This can be done without cairo, removing the need for it as an external dependency. (Angles and Fourways are still drawn with cairo, but Gtk.Snapshot provides the context).

~~If there is more than one keyboard device connected that might cause a problem for the code in this PR to detect ctrl/shift pressed with the mouse button on the fractal. PyGObject can't handle Gdk.Event types, which is needed to get the modifier from the drag gesture. Therefore, a workaround is used to get the modifier directly from the keyboard. Possibly it could be extended to cover the multi-keyboard situation.~~

When? GTK 4 is available in Ubuntu 22.04, a few months to go before a new release of Debian stable with it.

### DIFF
--- a/.github/workflows/test-gnofract4d.yml
+++ b/.github/workflows/test-gnofract4d.yml
@@ -41,7 +41,9 @@ jobs:
         # dbus-x11 avoids timeouts, failed to execute "dbus-launch"
         # gvfs is needed for Gio.AppInfo.launch_default_for_uri()
         # libxml2-utils provides xmllint for xml-stripblanks in gresource.xml
-        run: sudo apt install dbus-x11 gir1.2-gtk-4.0 gvfs libgirepository1.0-dev libxml2-utils xvfb
+        run: |
+          sudo apt update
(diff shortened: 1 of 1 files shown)


edyoung:
Thanks a lot Chris! Very exciting.

cjmayo:
There is a specific `gtk_event_controller_get_current_event_state` which can get the keyboard modifier from the gesture - don't need a workaround there.

Think that is it now, but I guess we'll wait a bit longer anyway.

cjmayo:
I haven't found any additional problems and Debian 12 is now stable, so I guess most people now have or have a route to GTK 4 (tests running fine here on macOS thanks to Homebrew). I propose to merge this next Monday (17th), do say if you think it is still too soon.

dragonmux:
Sounds good to us

PR #256, opened 2023-09-07
Add g++ and setuptools to dependencies in README.md

cjmayo:
I've done some testing on Debian and Ubuntu live DVDs and found g++ and python3-setuptools are not always included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,22 +36,24 @@ Gnofract 4D requires these packages to run:
 - Python version 3.6 or higher
 - GTK version 4.6 or higher
 - PyGObject version 3.29.2 or higher to access GTK from Python
-- A C++ compiler (used at runtime to compile your fractal formulas)
+- A C compiler (used at runtime to compile your fractal formulas)
 
-On Ubuntu, these can be installed with:
+On Debian/Ubuntu, these can be installed with:
 
     sudo apt install gcc gir1.2-gtk-4.0 python3-gi
 
 To build from source you also need:
 
+- A C++ compiler
 - headers for libpng and libjpeg
 - Python headers
 - glib-compile-resources and optionally xmllint
 - pkg-config
+- Python setuptools package
 
-On Ubuntu, these can be installed with:
+On Debian/Ubuntu, these can be installed with:
 
-    sudo apt install libglib2.0-dev-bin libjpeg-dev libpng-dev libpython3-dev libxml2-utils pkg-config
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,22 +36,24 @@ Gnofract 4D requires these packages to run:
...
 - Python headers
 - glib-compile-resources and optionally xmllint
 - pkg-config
+- Python setuptools package
 
-On Ubuntu, these can be installed with:
+On Debian/Ubuntu, these can be installed with:
 
-    sudo apt install libglib2.0-dev-bin libjpeg-dev libpng-dev libpython3-dev libxml2-utils pkg-config
+    sudo apt install g++ libglib2.0-dev-bin libjpeg-dev libpng-dev libpython3-dev libxml2-utils pkg-config python3-setuptools
```
dragonmux: Rather than directly installing g++, it is preferable to install `build-essential` so the user also gets all the core dependencies required for compiling and linking (eg, libc-dev, libc-bin-dev, etc).

cjmayo: `build-essential` which Debian describes as "packages which are considered essential for building Debian packages" installs:

dpkg-dev
g++
gcc
libc6-dev
make

dpkg-dev (which brings in fakeroot and a couple of Perl modules etc.)  and make we don't need, obviously gcc and g++, leaving only libc6-dev.

On Ubuntu installing gcc installs libc6-dev.

I think I'd rather avoid it, and if anything add libc6-dev for completeness.

dragonmux: There's a subtle reason why we're suggesting this meta package that isn't obvious from the packages it installs - and that's to do with installation status (automatic, vs manual) and package nomenclature changes (if one day, libc6-dev changes name to, say, libc7-dev, manual installation will break, but they guarantee that this meta package will continue to just work)

It might bring in more than we need, but it's safe, non-brittle, and doesn't wind up with the user being "stuck" with manual installation marked packages if ever they want to remove Gnofract4d and these packages after - meaning apt/dpkg's cleanup commands will work properly.

da2ce7: I concur with @dragonmux 's reasoning: we are not wanting a particular c++ compiler; we are wanting standardized build environment for gnofract4d to make use of.  Lets not micro-manage our users systems, but let it upto the distribution managers to select what exactly should be in a basic build environment.

cjmayo: OK updated, that makes sense. It was a rather experimental approach from the Gentoo using Python person!

dragonmux (approved):
LGTM now

PR #258, opened 2024-03-24
Fix building when pytest is not installed

### LINKED ISSUE #257: meson setup ERROR: Tried to use not-found external program as test exe
setup fails if pytest is not installed.

cjmayo:
Closes #257.

I've also raised the macOS version used in the workflow - gtk4 isn't available for macOS 11 from Homebrew, slowing the check.


dragonmux:
With macOS 11 Homebrew, it's slightly more that the person who heads the project withdrew all pre-built bottles for 11, so all `brew install` on 11 will build from source.

PR #260, opened 2024-03-25
Raise minimum Python version to 3.9, start testing 3.13

cjmayo:
PyEval_InitThreads() does nothing in Python 3.9-12 and is dropped in 3.13.

---

Python 3.8 does have a few months left but with the change to GTK 4 we dropped compatibility with Ubuntu 20.04 anyway.

### DIFF
--- a/.github/workflows/test-gnofract4d.yml
+++ b/.github/workflows/test-gnofract4d.yml
@@ -13,12 +13,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-22.04]
-        python: ["3.8", "3.9", "3.10", "3.11"]
+        python: ["3.10", "3.11", "3.12", "3.13-dev"]
         toxenv: [py]
         # Ensure ../codecov.yml uploads after all py jobs
         include:
           - os: macos-12
-            python: "3.7"
+            python: "3.9"
             toxenv: py
 
     env:
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Requirements
 
 Gnofract 4D requires these packages to run:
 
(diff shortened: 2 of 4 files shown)


cjmayo:
Looks like Meson 1.4.0 isn't compatible with Python 3.13 yet. Will investigate.

cjmayo:
A feature that Meson relies on has been removed. It wasn't listed for removal in 3.13 in the original Python 3.12 documentation, and looks like it is now coming back. Could be fixed in a couple of weeks time with the next Python 3.13 alpha. This was one that needed to sit for a bit anyway.

Perhaps it is a bit early to start testing with Python 3.13, but as the purpose was to be compatible with it...

No rush.

cjmayo:
Python 3.13.0a6 is released so Meson is happy.

But, in news elsewhere, Homebrew have updated to packages that do not work on macOS, causing a test failure:

```
 >       @pytest.mark.skipif(Gio.AppInfo.get_default_for_uri_scheme("http") is None,
                            reason="No web browser found")
E                           AttributeError: type object 'AppInfo' has no attribute 'get_default_for_uri_scheme'
```

https://github.com/Homebrew/homebrew-core/issues/167524
https://gitlab.gnome.org/GNOME/glib/-/issues/3322

Added a commit here for now to bypass that to see the Linux tests pass. I guess we wait for a new release to fix that, and we have the time.

dragonmux (approved):
Looks good to us - feel free to merge when you're ready to.

cjmayo:
Let's get on with it - the macOS job is causing all check runs to fail. Compilation on macOS is being run at least.

PR #263, opened 2024-08-20
Test on macOS 13

cjmayo:
GitHub is retiring their macOS 12 image.

Not an immediate hurry, 3 and a bit months yet:
https://github.blog/changelog/2024-08-19-notice-of-upcoming-deprecations-and-breaking-changes-in-github-actions-runners/


dragonmux (approved):
LGTM! Lucky that gnofract4d didn't get stuck on the macOS 11 image which got silently retired a couple of months back without warning. Keeps things much simpler.

PR #265, opened 2025-06-09
GitHub Actions Workflow Update

cjmayo:
No need to allow macOS to fail now because Gio has been fixed.
But Ubuntu 22.04 is not compatible with the latest PyGObject (and I get a test failure with Ubuntu 24.04 on GitHub that I can't fix - or replicate locally). Still testing the latest on macOS thanks to Homebrew.
And being able to manually trigger the workflow would have been useful to verify Gio was fixed without having to add any commits.


dragonmux (approved):
LGTM - changes all seem quite reasonable. Feel free to merge when you're ready.

PR #266, opened 2025-06-10
Test with Python 3.14

### LINKED ISSUE #267: Tests fail on Python 3.14
With Python 3.14, the `1 - 0j` iteration of the codegen tests started failing:

```
fract4d_compiler/tests/test_codegen.py:2305: in assertCSays
    self.assertEqual(exp, res)
E   AssertionError: 'ct_sin6 = (0.841471,-0)' != 'ct_sin6 = (0.841471,0)'
E   - ct_sin6 = (0.841471,-0)
E   ?                     -
E   + ct_sin6 = (0.841471,0)
```

This is due to a combination of `-0j` being handled differently in 3.14 together with the sign of `-0` not always being preserved.

More details:
https://github.com/python/cpython/issues/69639
https://bugzilla.redhat.com/show_bug.cgi?id=2339539

cjmayo:
With Python 3.13 and earlier:

```
>>> 1-0j
(1+0j)
```

With Python 3.14.0b2:

```
>>> 1-0j
(1-0j)
```

(is that better?)

resulting in:

```
fract4d_compiler/tests/test_codegen.py:2306: in assertCSays
    self.assertEqual(exp, res)
E   AssertionError: 'ct_sin6 = (0.841471,-0)' != 'ct_sin6 = (0.841471,0)'
E   - ct_sin6 = (0.841471,-0)
E   ?                     -
E   + ct_sin6 = (0.841471,0)
```

When investigating that I found that manufacture_tests() was missing off the last test.


cjmayo:
Breaks:

```
msg = "'ct_cosxx10 = (0.540302,-0)' != 'ct_cosxx10 = (0.540302,0)'\n- ct_cosxx10 = (0.540302,-0)\n?                        -\n+ ct_cosxx10 = (0.540302,0)\n"

    def fail(self, msg=None):
        """Fail immediately, with the given message."""
>       raise self.failureException(msg)
E       AssertionError: 'ct_cosxx10 = (0.540302,-0)' != 'ct_cosxx10 = (0.540302,0)'
E       - ct_cosxx10 = (0.540302,-0)
E       ?                        -
E       + ct_cosxx10 = (0.540302,0)
```

dragonmux:
Looks like it might be worth it to special-case assertEqual there for signed zero and consider -0 == 0 for the purposes - given this all stems from how floating point handles this in the IEEE spec in the first place, this isn't otherwise fixable we would imagine. The old behaviour of considering `-0j` to be `+0j` in conversion was wrong from the IEEE perspective, but we don't think that should mean this should be a test failure seeing that behaviour corrected..

cjmayo:
The assertEqual is comparing the strings though...

Adding the `-1 + 0j` is causing failures with Python < 3.14 - splitting that out: #268

cjmayo:
The string replace from https://github.com/fract4d/gnofract4d/pull/268 does make current master tests pass though.

cjmayo:
OK, now with string replacement.

dragonmux (approved):
Ah! had somehow missed this was string comparisons. The suggested fix looks good to us though so go for it 🙂

PR #271, opened 2025-06-15
Replace get_allocated_height/width

cjmayo:
None of the GTK deprecation warnings actually matter at this point, but these ones are easy and save a few characters.

https://docs.gtk.org/gtk4/method.Widget.get_allocated_width.html


dragonmux (approved):
LGTM, feel free to merge this when you are ready.

PR #274, opened 2025-06-16
Add a workflow to create a branch with updated help

cjmayo:
Some minor fixes to the manual source (which is used for application help and the website).

And a GitHub workflow that can generate the help folder in a new branch that can be submitted as a PR.

This way you don't have to have the correct version of hugo installed etc.


dragonmux (approved):
This LGTM, feel free to merge it when you're ready.

cjmayo:
Corrected one line:

```
git commit -a -m "Update help"
```

to

```
git add help/
git commit -m "Update help"
```

because as hugo versions change looks like the css (which contains the hash in its filename) can change.

PR #276, opened 2025-06-16
Create release file with a GitHub Actions workflow

cjmayo:
This workflow automatically creates and adds a release tar.gz to the GitHub release announcement.

It regenerates help because the version is used in there.
It uses actions/attest-build-provenance to automatically sign the tar.gz (shown on the Action not the release...)


.github/workflows/release-file.yml
```diff
@@ -0,0 +1,50 @@
...
+  upload_archive:
+    name: Upload archive
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+      attestations: write
+    steps:
+      - run: |
+          ref=${{ github.ref_name }}
```
dragonmux: We would suggest enclosing the expansion here in double quotes, as a precaution. ie `ref="${{ github.ref_name }}"` - that way if a ref is created that would otherwise cause extra assignments or insert spaces here, it won't cause problems.

.github/workflows/release-file.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          path: ${{ env.release_name }}
+          submodules: recursive
+      - name: Install Ubuntu packages
+        run: sudo apt install hugo
+      - name: Save version
+        run: >
+            echo ${{ github.ref_name }} > ${{ env.release_name }}/.gnofract4d_version.txt
```
dragonmux: Similar points for these - just as a guard.

dragonmux (commented):
There are a couple of things we'd tweak, but otherwise this looks good.

.github/workflows/release-file.yml
cjmayo: Thanks. Added in a new commit I'll include.

.github/workflows/release-file.yml
cjmayo: OK. I think I've done them all. I'll retest this at some point.

dragonmux (approved):
LGTM, feel free to merge this when ready.

cjmayo:
Still works!

PR #279, opened 2025-06-19
Add doctype and lang to website

cjmayo:
Avoid browsers using quirks mode.


dragonmux (approved):
LGTM - good upgrade as quirks mode tends to cause massive headaches. Feel free to merge this when ready.

PR #282, opened 2025-07-15
Document a release process

cjmayo:
I think it would be reasonable to create a 4.4 release and collect all the changes of the last few years. (I'm going to have to package something new for Gentoo anyway to get the tests passing for Python 3.14).

The recent changes have had a bit of time now.

Here's what I think the steps should be.


dragonmux (approved):
This looks reasonable to us. We agree that there's not been a release in a bit long, so cranking the handle is a little overdue. Go ham!

cjmayo:
Thanks.

PR #283, opened 2025-07-18
Fix Hugo generated help dir attribute value

cjmayo:
Hugo 0.123.7 creates:

`<html lang="en" dir=ZgotmplZ>`

`<body dir=ZgotmplZ>`

Also languageCode = "en-us" is ignored.

----

Release step 2 aborted.


dragonmux (approved):
Given the description, LGTM - hopefully this makes Hugo do something a bit more sensible. Feel free to merge when ready.

PR #285, opened 2025-07-22
Reduce release file size

cjmayo:
Don't install documentation sources and use xz compression.

---

The 4.3 file was 866 KB, we were at 11MB! This gets it down to 2.62 MB, all testdata is still included..

The gz automatically generated by GitHub will still be available if either of these changes are a problem to people.

(Looks like a small problem with byte-compiling, so no release just yet).


cjmayo:
OK. They were being added to the file - remove them - now 2.57MB.

dragonmux (approved):
LGTM! Feel free to merge this when ready.

cjmayo:
Thanks. I decided to not create the `.pyc` files with `python -B createdocs.py` instead of deleting them afterwards.

PR #286, opened 2025-07-23
Update theme used for the manual to hugo-book v9

cjmayo:
The "ltr" problem was in the theme (v7).

I've added branches to https://github.com/fract4d/hugo-book with the customised versions.


dragonmux (approved):
LGTM, feel free to merge when ready. Good sleuthing work on why this broke, and better fixing it!

cjmayo:
Thanks. For the record the current theme was actually v7 + a couple of months, almost v8.

